PR #36, opened 2020-10-02
Droid Merchant

OrehRuoy:
Droid Merchant. Also changes Spaceport Speeders to enable Droid Merchant to work with it.


gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set1/light/Card1_112.java
```diff
@@ -32,7 +32,7 @@ public Card1_112() {
 
     @Override
     protected List<PlayInterruptAction> getGameTextTopLevelActions(final String playerId, final SwccgGame game, final PhysicalCard self) {
-        Filter filter = Filters.and(Filters.your(self), Filters.or(Filters.droid, Filters.vehicle), Filters.at(Filters.or(Filters.Mos_Eisley, Filters.sameSiteAs(self, Filters.Wioslea))));
+        Filter filter = Filters.and(Filters.your(self), Filters.or(Filters.droid, Filters.vehicle), Filters.at(Filters.or(Filters.Mos_Eisley, Filters.sameSiteAs(self, Filters.Wioslea),Filters.sameSiteAs(self, Filters.Droid_Merchant))));
```
jdipaolo1, line 35: You'll want to accomplish via a gameTextModificationModifier. Look at Kalit for an example. That way if Droid Merchant's gametext is cancelled this will no longer be in affect.

PR #186, opened 2020-10-27
Fix for HCF/HCFv and Protector

kpbabb:
Adds Han and Chewie personas to HCF/HCFv while card is in hand so that Protector can be used with HCF/HCFv


fastlearner-ai:
This same fix will probably have to be made for Death Star Assault Squadron to get the new I've Been Searching For You For Some Time (V) card to work (Lose Vader from hand...)

EchoBaseTrooper:
I think a better solution would be to make GEMP check the persona of permanent pilots/astromechs/etc. if it's not on the table. Adding the persona to the card will make it work incorrectly in certain situations.

PR #235, opened 2020-11-24
Update Card2_023.java

Solidsnack3:
Adjusting A New Hope Wedge to add his character modifiers to the persona of Red_2 so he can interact properly with Red Squadron 1


EchoBaseTrooper:
He already worked correctly, right? Did you test to see if there was a bug?

PR #282, opened 2021-01-19
Fix/lr fix wrong card  on ban list

stevetotheizz0:
Fixed ban list Sim Aloo (v) 203_15 was accidentally added to the format instead of TSID (v) 205_15


EchoBaseTrooper:
Made changes in #283

PR #287, opened 2021-01-25
Fixing Issue#38 - 'Concentrate All Fire' & 'Fighter Cover' Now only apply the +3 power bonus to a ship that's fired a weapon once per battle

### LINKED ISSUE #38: Concentrate All Fire + B-Wings
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=jokerking$r2bztq83l0oaeqbv

LS players was able to use Concentrate All Fire to add to the power of the same starfighter multiple times in the same battle. Based on the ruling below, that should not happen.

viewtopic.php?f=5&p=1310843#p1310843

### LINKED ISSUE #46: Fighter Cover Accumulation (TIE INTERCEPTOR)
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=dvphimself$0u6tot800wl1i68j

Hello, this in game event has been reviewed by Gergall and Death Star in the rules thread, and has been confirmed as a bug.

firing an SFS Cannon twice with a Tie Interceptor will reliably add the 3 in response to both firings.

Relevant Ruling
https://forum.starwarsccg.org/viewtopic.php?f=5&t=75413

Ten19:
Raised in Issue #38 , corrected for both 'Concentrate All Fire' & 'Fighter Cover'. Both cards now check if the ship firing the weapon already has power modifier supplied by the Admiral's Order, and does not trigger if said modifier exists.
Closes #38 
Closes #46

### DIFF
--- a/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set9/dark/Card9_093.java
+++ b/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set9/dark/Card9_093.java
@@ -23,6 +23,7 @@ import com.gempukku.swccgo.logic.timing.EffectResult;
 import com.gempukku.swccgo.logic.timing.results.FiredWeaponResult;
 
 import java.util.Collections;
+import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


EchoBaseTrooper:
This does make this specific interaction work correctly, but will potentially leave a problem with the cumulative rule with other cards. Fixing the power calculation to correctly apply the cumulative rule seems like it would be a better solution.

BOTVHD:
@EchoBaseTrooper does it make sense to implement this patch to these 2 specific cards and open a separate issue for the broader cumulative rule issue (not preventing one card from applying modifiers multiple times on the same card)?

The new issue would reference these 2 patches that can be removed once the cumulative rule is addressed at a system level.  Similar to the way Luke Skywalker, The Last Jedi has a patch/fix until the system level fix is implemented ( https://github.com/PlayersCommittee/gemp-swccg-public/issues/890 )

I only suggest this approach because, per comments here https://github.com/PlayersCommittee/gemp-swccg-public/issues/38#issuecomment-3481602002 it feels like this issue is more impactful than most of bugs and it could take a while before the cumulative rule is fixed?

PR #532, opened 2022-03-18
update vbox static ip address

jetsaredim:
Newer versions of VirtualBox restrict the static ip range for hostonly networks to between 192.168.56.1 and 192.168.63.254.  This picks a nice static ip in the middle.


DevoKun (approved):

DevoKun (changes requested):
Please also update the `README.md` to use the new IP address: `192.168.50.94`

DevoKun (approved):

PR #657, opened 2023-01-02
Remove files from errata

jetsaredim:
Macroscan was removed from set 10 during the most recent errata.  The sync from the private repo doesn't remove files so it needs to be cleaned up manually.


jetsaredim:
need to re-base on master

PR #740, opened 2024-02-09
Docker Volume Bind for Web Application

nheinbaugh:
Minimal changes required to bind the host `gemp-swccg-async/src/web/` folder to the correct location in the docker container. This makes modifications to the web app easier to develop and test by removing the manual steps required to rebuild the application with maven, zip and replace the folder in the container and refresh. 

### Implementation Steps
* A new environment variable defines if it is a prod or test build 
  * This defaults to prod to mimic previous behavior. Since there was no flag everything was a "prod" build
  * The `Dockerfile` where "default parameters" for a "test" environment is updated
* The two HTTP handlers both have a property that indicates if the app is local or prod
  * I don't know enough about Java or the app to find a way to inject this configuration. I do hate logic in constructors... but this seems reasonable
* Updated `Dockerfile` to no longer copy the web folder manually
* Updated the `docker-compose.yml` file to bind the host web folder into the container

### Testing
* I validated the current behavior for making any changes in an HTML or JS file were as follows:
  *  Make change
  * Refresh page in UI
    * See the changes are not represented
  * Stop the docker containers
    * Following a restart the the changes are still not represented
  * Delete the docker image
    * Following a deletion of the container and then the image, `docker-compose up` and the changes are not represented
  * Run `mvn clean install`
    * After deleting the container, the image and then rebuilding the changes are finally represented

* The current flow is now:

* With a `local` build
  * Make a change in any HTML or JS file
  * Refresh the page in the browser
  * See changes are represented

* With a `production` build
  * Make a change in any HTML or JS file
  * Refresh the page in the browser
    * See no changes are visible
    * See the requests all return `304` (not modified)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,9 @@ RUN mkdir -p /opt/gemp-swccg/web && \
 ## a small quality of life addition for those times when entering the container
 RUN echo 'export PS1="\u@\h:\w\$ "' > /etc/profile.d/ps1.sh && chmod +x /etc/profile.d/ps1.sh
 
+COPY gemp-swccg-async/src/main/web \
+     /opt/gemp-swccg/web
+
 COPY gemp-swccg-async/target/web.jar \
      /opt/gemp-swccg/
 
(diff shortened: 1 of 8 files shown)


ajhat16 (approved):

jdipaolo1 (approved):

jetsaredim:
this should be closed as it was superseded by changes made in the private repo.

Issue #477, opened 2021-12-15
Major Rhymer - Opponents Using Landspeed to Run Away While Rhymer making a Bombing Run


Solidsnack3:
- [ ] Did you search through the list of **open issues** _first_ to ensure that there is not already an report for this issue?
-Yes

## **Describe the bug**
- Super low priority I assume, but wanted to get it out there.  Major Rhymer's text prevents landspeed movement from opponent's characters at same site while making a bombing run aboard a bomber.  General Solo clearly moves away, using personal landspeed, while Rhymer is 'making a bombing run' at the back door.  From my understanding, Rhymer should prevent movement the entire time he is at that site, during the move phase and even during subsequent battles on the following turn.

## **To Reproduce**
1. heh.. bombing run has quite a few steps...
2. Deploy a system and two related sites
3. Deploy a Tie bomber with pilot capacity (Scimitar 1 or Scimitar 2) with Major Rhymer aboard
4. Have opponent deploy a character to one of the sites
5. deploy bombing run at the same site as opponent's character
6. move Major Rhymer's tie to the same site as opposing character using bombing run
7. see if opposing character can move away or not using personal landspeed

## **Game Replay ID:**
- Dark Side, Turn 10:
- https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=Solidsnck$760o99f6k4dy6kvq

## **Expected behavior**
-Major Rhymer, while aboard a Tie Bomber at a site, prevents opponents from moving away using landspeed.

Gergall:
Noobi-wan mentioned to me that this bug happened to him recently.

billbisco:
@Gergall    This Bug happened to me as well.  I confirm.  I also have my own replay:  https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=billbiscos$wiziu1jhv07cfbqn

If you go to turn 12, you will notice that when a battle is initiated against a lone Obi-Wan, Obi-Wan can use Dodge as well.  

Major Rhymer doesn't stop regular movement nor react movement.  It also should not allow Luke to move using Run Luke Run either.

BOTVHD:
Major Rhymer: Card9_115

Currently coded with this modifier to apply the limitation:
```
        modifiers.add(new MayNotMoveUsingLandspeedModifier(self, Filters.and(Filters.opponents(self), Filters.character, Filters.atSameSite(self)),
[...]

PR #912, opened 2025-12-31
Fix missing ModifierType enum for landspeed movement restriction

Rad10Snack:
This fixes a compilation error in the public repo where
MayNotMoveUsingLandspeedModifier references a missing ModifierType enum value.

The modifier class already exists and is referenced by card logic,
but the corresponding enum


BOTVHD:
@Gergall does it make sense that the public repo wouldn't compile?  I thought it was basically a ported (compiling) snapshot that matched what is running live on the gemp server?

Gergall (changes requested):
There is an issue, but we need a different fix to match the private repo.

SHORT ANSWER: The correct PR is to delete MayNotMoveUsingLandspeedModifier.java which is no longer used by any card and never will be.
Please change this PR to undo the current change, and to delete \src\gemp-swccg-logic\src\main\java\com\gempukku\swccgo\logic\modifiers\MayNotMoveUsingLandspeedModifier.java

LONGER ANSWER: The file was intentionally deleted from the private repo after fixing a bug with Major Rhymer. Rhymer and a couple other cards use "MayNotMoveFromLocationToLocationUsingLandspeedModifier.java" and many cards use "ResetLandspeedModifier.java" to reset landspeed = 0.
Public info on the Rhymer fix: https://github.com/PlayersCommittee/gemp-swccg-public/issues/477
Private info on the Rhymer fix: https://github.com/PlayersCommittee/gemp-swccg/pull/3012

Apparently, deleting the file from the private repo did not automatically "flow" to the public repo. So it should be deleted manually.  Besides that, the file was bugged (did not work at all, hence why Rhymer needed to be fixed).

rad10snack:
@Gergall Removed the java file and reverted the enumeration addition. Thanks!

Gergall (approved):

Gergall:
All set. Thank you for the fix.

PR #1004, opened 2026-08-28
Fix Wise Advice optional Immediate Effect free deploy (issue 990)

### LINKED ISSUE #990: Wise Advice / Oppressive Enforcement do not use 'may' for free Immediate Effects

## **Describe the bug**
Wise Advice and Oppressive Enforcement (total 6x cards from Effects, Defensive Shields, and combo cards) have "your Immediate Effects **may** deploy for free" which is different from "your Immediate Effects deploy for free".

When one of these cards is on table and the user wants to deploy an Immediate Effect, they should have the option to pay the normal cost (not using the 'may' text) or the modified cost (using the 'may' text to change it to 'free').

Currently, GEMP always applies the free modifier.

## **To Reproduce**
1. Have one of these cards in play (ex: Wise Advice)
2. Have an Immediate Effect in hand (ex: Grappling Hook)
3. Have sufficient force available to pay the 'normal' Immediate Effect cost (ex: 1 force available for Grappling Hook)
4. Attempt to play Immediate Effect (ex: Grappling Hook)
5. Notice that Immediate Effect plays free and no option is provided to play for normal cost (ex: 1 force for Grappling Hook)

## **Game Replay ID:**
N/A

## **Expected behavior**
Expect that with sufficient force available, 2 actions should be given instead of 1 for the Immediate Effect (a 'free' version and a 'normal' version of the action).

## **Screenshots**
N/A

## Client Platform
N/A

## **Additional context**
This is very similar to the way Battle Plan says "you **may** initiate battles for free" and in GEMP, this is implemented by providing the user with two separate initiation actions (with sufficient force).

billbisco:
Why: leftover #990. Wise Advice / Oppressive Enforcement (and the combo / Defensive Shield versions) say Immediate Effects may deploy for free. GEMP was applying DeploysFreeModifier as always-free, so Grappling Hook only offered the free grab and could not pay printed cost. Energy Walls on Generator Core is already 0 Force; Wise Advice must not add a second free / 0-Force copy of that same control-phase deploy.

What it Does: Switches six cards from DeploysFreeModifier to MayDeployFreeModifier (MAY_DEPLOY_FOR_FREE). AbstractImmediateEffect offers both free and printed-cost play actions (Battle Plan pattern; grabbers keep targeting). Free action is listed first; paid action labels the Force cost. Skips the extra free copy when the Immediate Effect already deploysForFree. Adds Card_7_081_Tests and Card_7_234_Tests, including WiseAdviceDoesNotDuplicateAlreadyFreeEnergyWalls.

What it Fixes: Grappling Hook (and other Immediate Effects) can still pay printed cost with Wise Advice on table. Energy Walls on Generator Core stays a single 0-Force control-phase deploy.

Issues: Closes #990


BOTVHD:
How does this interact with Immediate Effects that already have no cost?  Ex: playing Energy Walls on Generator Core with Wise Advice on table.

BOTVHD:
I don't know if this is the right approach or not, but AbstractInterrupt does something kind of like this - where it merges game text card actions as well as granted actions and more:
```
        if (checkPlayRequirements(playerId, game, self, null, null, null)) {
            List<PlayInterruptAction> actionList1 = getGameTextTopLevelActions(playerId, game, self);
            if (actionList1 != null) {
                actions.addAll(actionList1);
            }

            // Check if interrupt was granted ability to perform a top-level action from another card
            List<PlayCardAction> actionList2 = getGrantedTopLevelActions(playerId, game, self);
            if (actionList2 != null) {
                actions.addAll(actionList2);
            }
        }
```
I wonder if the AbstractImmediateEffect changes proposed here should be reworked (either at the AbstractImmediateEffect level or higher up the chain) and have cards like Wise Advice grant a free action that copies the card's original action set, changes the cost to free, and appends the 'for free' text.  By ordering the granted actions group first, I assume you wouldn't even need to re-sort the list.

Maybe in getGrantedTopLevelActions (or its equivalent), you would look for the may deploy free modifier on the card. If it exists (and deploy was not already free), then construct the free copy of any deploy actions.

billbisco:
Energy Walls on Generator Core is already 0 Force. Wise Advice must not add a second free / 0-Force copy of that same control-phase deploy. The extra action is now skipped when the Immediate Effect already deploys for free. Grappling Hook still gets free + printed cost.

Grabbers like Grappling Hook are optional-before plays, not top-level granted actions, so the dual play actions stay on AbstractImmediateEffect the same way Battle Plan offers two initiate-battle actions. Thanks VHD.

PR #1006, opened 2026-08-28
Implement Targeting Computer (1_39)

### LINKED ISSUE #70: Targeting Computer
#

billbisco:
Implements Targeting Computer (`1_39`).

## Summary
Premiere Light Device: deploy for 2 Force on your starship (+1 maneuver). During battle, once per copy, fire a weapon twice as one overall action — Separately or Combined. Subtract 1 from each destiny draw only while using Fire a weapon twice.

## Card
- **Id / title:** `1_39` / Targeting Computer
- **Type:** Light Device, Premiere
- **Game text:** Use 2 Force to deploy on any of your starships. Adds 1 to maneuver. During battle, may fire one of that starship's weapons twice (as one action, separately or combined). Subtract 1 from each destiny draw while doing so.

## What changed
- Adds `Card1_039` with Separately / Combined / Don't Fire (Cancel) three-button UI
- Engine: `SeparatelyOrCombinedFiringState` plus hooks in `GameState`, `WeaponFiringState`, `DrawDestinyEffect`, and `Destiny` so combined shots skip per-firing total modifiers and resolve after both draws
- Separately: choose/resolve, then choose/resolve again (may retarget); Force paid per firing as each shot initiates
- Combined: each firing is a single weapon destiny (draw modifiers only); destinies added, then total weapon destiny modifiers once vs one target
- Once per copy per turn (not merely once per battle); usage recorded when Separately or Combined is chosen; Don't Fire leaves the copy unused
- Device-per-turn slots: starfighter 1, squadron 3, capital unlimited
- Per-turn device/weapon slot counting uses **unique** copies, so firing a weapon twice still spends one weapon slot and one device slot (a squadron can still use three different Targeting Computers in one turn)
- Crash/leak fixes: finish separately-or-combined firing state only after a shot has started; unused attached TC must not −1 ordinary Fire; fire-twice −1 must still apply when SOC is live

## Design choices
- Not Maul-style repeated firing — one overall action with Separately or Combined
- Appendix B option A: if the second shot cannot pay, the first still resolved when separately; completed combined firings still combine
- Multiple Targeting Computers on the same ship do **not** stack +1 maneuver (Cumulative Rule)
- Continuous +1 maneuver is not “using” the device
- Combined Attack / Precise Attack are not in this PR (see #1007)

## Tests
`Card_1_39_Tests` — **37** tests:
- Stats; deploy 2 Force; +1 maneuver; cannot use outside battle
- One usage per device; fire twice separately (retarget, Force twice); combined destinies then total mods
- TC −1 each draw (not once on combined total); combined can hit when separate would miss
- XWLC / SW-4 Ion Cannon separately and combined; XWLC at X=3 combined loses the TIE
- Unpaid second separately does not rewind first; combined unpaid shot-2 still applies completed firings
- Don't Fire leaves TC unused; HTB from Home One separately/combined
- Defiance + HTB +2 each draw with/without TC; SOC cleared after separately + Defiance response (no snapshot crash)
- Playtest −1 path: Defiance+Karie+Dack+Verrack+HTB vs Executor separately → 8 and 8 (not 9)
- Unused TC does not −1 ordinary Fire (Defiance+Verrack+HTB vs Executor: 7 and 10)
- Falcon/CEC/Karie/Rogue Gunner/QLC and Ten Numb/Blue 5/Concussion Missiles EACH vs TOTAL coverage
- Four copies on B-wing: maneuver +1 not +4; starfighter/squadron/capital per-turn copy limits

## Known gaps
- Combined Attack / Precise Attack live on later stacked PRs (#1007 / #1014)

Closes #70

---

### DIFF
--- a/src/gemp-swccg-async/src/main/web/js/gemp-016/gameUi.js
+++ b/src/gemp-swccg-async/src/main/web/js/gemp-016/gameUi.js
@@ -2405,12 +2405,13 @@ var GempSwccgGameUI = Class.extend({
         var text = decision.getAttribute("text");
         var results = this.getDecisionParameters(decision, "results");
         var defaultIndex = this.getDecisionParameter(decision, "defaultIndex");
+        var asButtons = this.getDecisionParameter(decision, "asButtons") == "true";
 
         var that = this;
         this.smallDialog
             .html(text);
 
-        if (results.length > 2 || this.settingsAlwaysDropDown || defaultIndex >= 0) {
+        if (!asButtons && (results.length > 2 || this.settingsAlwaysDropDown || defaultIndex >= 0)) {
             var html = "<br /><select id='multipleChoiceDecision'>";
             for (var i = 0; i < results.length; i++) {
                 html += "<option " + (i == defaultIndex ? "selected='selected' " : "") + "value='" + i + "'>" + results[i] + "</option>";
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set1/light/Card1_039.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set1/light/Card1_039.java
@@ -215,7 +215,7 @@ public class Card1_039 extends AbstractDevice {
                 game.getGameState(), soc.getCardFiringWeapon(), weapon, null,
                 Collections.singletonList(target), combined);
         soc.markResolved();
-        game.getGameState().sendMessage("Combined total weapon destiny (completed firings only): " + GuiUtils.formatAsString(totalDestiny));
+        game.getGameState().sendMessage(soc.getCombinedTotalDestinyMessage(GuiUtils.formatAsString(totalDestiny)));
         float defenseValue = game.getModifiersQuerying().getDefenseValue(game.getGameState(), target);
         game.getGameState().sendMessage("Defense value: " + GuiUtils.formatAsString(defenseValue));
         if (totalDestiny > defenseValue) {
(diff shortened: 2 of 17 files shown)


ketura:
1- In the test cases, there are a number of private functions at the bottom; some of these are particular to the test case here and are fine to have as-is, but a few of them incorporate functions that should be included in the greater test rig so they can be accessed by other tests (since they are general-purpose functionality that many tests can benefit from):

- startWeaponsSegment (as: StartBattleAndSkipToWeaponsSegment)
- ensureLSForcePile (with DS versions as well)
- drainLSForcePileTo (with DS versions as well)
- isIonized

2- passOptionalResponses already exists; use scn.PassAllResponses() instead.

3- I will also express that chooseWeaponIfPrompted is not a good pattern to follow in these tests: since you have control of the test, you get to decide if there *is* a prompt or not.  If there isn't, you shouldn't be blindly checking, and if there is, you should be handling it.  Skipping over the standard optional response spam is understandable, but from the perspective of having robust tests, they should know their context and enforce them, not let them possibly-work.  

The idea is that if somewhere along the line the basic assumptions made by these tests are violated, then they *break*, loudly and obviously, and having a check-and-handle function like that introduces a flexibility to the system that it shouldn't have.  They should break, not bend.

Other than that, these are decent tests.  I will say that this particular card more than any other has potential of nearly as many edge cases as there are weapons in the game, and so it is quite likely that it may require that some exhaustive tests be added to check the basic assumptions of combat, to ensure this card's alterations A: don't break those undocumented assumptions and B: this card works with those weapons.

This effort was one that we had all assumed would require a complete rewrite of the weapon system, and essentially every weapon in the game as a result. I am surprised that a solution could be found that could be hacked in so seemingly easy, but I bet it's going to require a relative shitload of weapon + combat tests (in the rules folder of the test cases) to build up trust in the solution.

billbisco:
@ketura   Thank you very much for your kind words and feedback.  Some changes were made in response.  Let me know what you think about the current state and anything else missed and can improve.  And, if I missed something you said to improve earlier, please kindly remind me again.     Playtesting these cards is very exciting!

PR #1009, opened 2026-08-30
Stay Sharp: errata gametext/JSON and total weapon destiny stacking fix

billbisco:
Two Stay Sharp! fixes: Advanced Rulebook errata text (Java + JSON), and a bug where the +2 could stack onto each Heavy Turbolaser Battery destiny draw instead of applying once to the total weapon destiny.

## 1. Errata gametext
GEMP Java `setGameText` and JSON (`7_104` in `card_blueprint_database.json` and `card_blueprint_database_light.json`) were still on the old wording (*may add 2 to destiny draw*). They now match the [2023 Advanced Rulebook](https://res.starwarsccg.org/rules/SWCCG_2023_AdvancedRulebook.pdf):

> During your control phase, fire one of your starship weapons (for free). If Han or any gunner is aboard that starship, may add 2 to the total weapon destiny. 'Hit' target is lost. OR If you just fired a weapon in battle, add that weapon's destiny number to your total power.

## 2. Total weapon destiny stacking bug
Heavy Turbolaser Battery draws two destinies. Stay Sharp was using `ModifyDestinyEffect` (per-draw), so the optional "Add 2" after each draw could stack to +4. The errata adds 2 to the **total** weapon destiny once.

- Optional is still offered after each weapon destiny draw of that Stay Sharp firing until it is accepted once. Declining the first draw still offers it on the second (or on a Concentrate All Fire redraw). Accepting it does not offer it again, and the same bonus cannot stack to +4.
- The +2 is applied with `ModifyTotalWeaponDestinyBeforeDrawingDestinyEffect` (until-end-of-weapon-firing `TotalWeaponDestinyModifier`), so it survives a canceled/redrawn individual destiny.
- Han/gunner is checked on the firing starship at fire time (Stay Sharp is already off table).
- Card-level only (`Card7_104`). No engine cumulative changes, and no Targeting Computer / Combined Attack / Local Trouble / Frustration edits.

## Tests
`Card_7_104_Tests` (5 tests, 0 fail via docker Maven). Latest: [`c63adef`](https://github.com/PlayersCommittee/gemp-swccg-public/commit/c63adef74e3c357cb3b2788a07e35b47f52e3f3d).

- [x] `StaySharpStatsAndKeywordsAreCorrect`
- [x] `StaySharpHeavyTurbolaserBatteryCannotStackPlusFourToHitDuringControlPhase` — take +2 whenever offered. Destinies 1 then 3 vs Imperial-Class Star Destroyer armor 6: base `1+3-1=3`; +2 = 5 miss; +4 = 7 would hit (`>` defense). Assert the Destroyer is **not** hit.
- [x] `StaySharpPlusTwoStillOfferedOnSecondDrawIfDeclinedOnFirst` — decline on draw 1, accept on draw 2, still only +2 (same miss).
- [x] `StaySharpPlusTwoSurvivesConcentrateAllFireRedrawAfterAccept` — accept Stay Sharp +2 after the first draw, then Concentrate All Fire (`9_3`) cancels and redraws that destiny. Final destinies 2 then 4: `2+4-1=5` miss without the bonus, `7` hits armor 6. Assert hit, offer-once, no +4 stack.
- [x] `StaySharpStillOfferedAfterConcentrateAllFireRedrawIfNotYetAccepted` — skip Stay Sharp by taking Concentrate All Fire first; Stay Sharp is still offered on the redraw and accepted once; same hit math.

---

### DIFF
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set7/light/Card7_104.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set7/light/Card7_104.java
@@ -20,8 +20,8 @@ import com.gempukku.swccgo.logic.actions.OptionalGameTextTriggerAction;
 import com.gempukku.swccgo.logic.actions.PlayInterruptAction;
 import com.gempukku.swccgo.logic.actions.TriggerAction;
 import com.gempukku.swccgo.logic.effects.FireWeaponEffect;
-import com.gempukku.swccgo.logic.effects.ModifyDestinyEffect;
 import com.gempukku.swccgo.logic.effects.ModifyTotalPowerUntilEndOfBattleEffect;
+import com.gempukku.swccgo.logic.effects.ModifyTotalWeaponDestinyBeforeDrawingDestinyEffect;
 import com.gempukku.swccgo.logic.effects.RespondablePlayCardEffect;
 import com.gempukku.swccgo.logic.effects.choose.ChooseCardOnTableEffect;
 import com.gempukku.swccgo.logic.timing.Action;
@@ -44,7 +44,7 @@ public class Card7_104 extends AbstractUsedInterrupt {
     public Card7_104() {
         super(Side.LIGHT, 4, "Stay Sharp!", Uniqueness.UNIQUE, ExpansionSet.SPECIAL_EDITION, Rarity.U);
         setLore("'Ha haaaaaa!'");
-        setGameText("During your control phase, fire one of your starship weapons (for free). If Han or any gunner is aboard that starship, may add 2 to destiny draw. 'Hit' target is lost. OR If you just fired a weapon in battle, add that weapon's destiny number to your total power.");
+        setGameText("During your control phase, fire one of your starship weapons (for free). If Han or any gunner is aboard that starship, may add 2 to the total weapon destiny. 'Hit' target is lost. OR If you just fired a weapon in battle, add that weapon's destiny number to your total power.");
         addIcons(Icon.SPECIAL_EDITION);
     }
 
@@ -105,7 +105,8 @@ public class Card7_104 extends AbstractUsedInterrupt {
                                                                     public List<TriggerAction> getOptionalAfterTriggers(String playerId3, SwccgGame game, EffectResult effectResult) {
                                                                         List<TriggerAction> actions = new LinkedList<TriggerAction>();
                                                                         // Offer after each weapon destiny draw until the player accepts once.
-                                                                        // Accepting adds 2 to that draw (and thus the weapon destiny total).
+                                                                        // Accepting adds 2 to the TOTAL weapon destiny for the rest of this firing
+                                                                        // (until-end-of-weapon-firing), so a cancel/redraw keeps the bonus.
                                                                         // Declining leaves it available on a later draw of this same firing.
                                                                         // The same Stay Sharp bonus does not stack if it is somehow accepted twice.
                                                                         if (playerId.equals(playerId3)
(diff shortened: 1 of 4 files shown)


BOTVHD:
Hmm, there is a problem with this - but I'm not sure this is the right solution.  It looks like this is modifying an individual weapon destiny (and your proposed changes are preventing that from happening more than once), but the errata game text says it is supposed to add to the **total** weapon destiny.

It seems like ModifyTotalWeaponDestinyEffect or ModifyTotalWeaponDestinyBeforeDrawingDestinyEffect should be getting used instead of ModifyDestinyEffect.

For example if an individual weapon destiny is canceled or redrawn, you want that total destiny modifier to stay intact.

I could be way off, but I'm guessing someone originally copied the card code from Sorry About The Mess (which does allow Han to add 1 to each weapon destiny - not the total weapon destiny!).

BOTVHD:
Also, it looks like there was an errata and some of the wording around adding to destiny was clarified.  See the AR entry and errata card image on scomplink.

The gemp card implementation's gametext (and currently, the json text) are out of date with this errata.

billbisco:
Switched Stay Sharp to a total weapon destiny modifier as suggested.

Used `ModifyTotalWeaponDestinyBeforeDrawingDestinyEffect` rather than `ModifyTotalWeaponDestinyEffect`. Despite the name, it only needs `WeaponFiringState` and adds `TotalWeaponDestinyModifier` via `addUntilEndOfWeaponFiringModifier`, so the +2 survives Concentrate All Fire cancel/redraw. The until-end-of-draw-destiny path would be wiped with the canceled draw.

Offer-until-accepted is unchanged: optional after each weapon destiny of that firing until taken once; decline on draw 1 still offers on draw 2 or on a redraw; accept once and it is not offered again and cannot stack to +4.

Also updated Java `setGameText` and the `7_104` JSON entries to the Advanced Rulebook wording (add 2 to the total weapon destiny). Added Concentrate All Fire redraw coverage in `Card_7_104_Tests`.

BOTVHD:
I'm not 100% sure how this repo interacts, but I think the JSON repo is entirely separate (https://github.com/swccgpc/swccg-card-json).  @Gergall has made a note about fixing that in a future json update.

I'm not sure about the timing for when you should make the choice about adding to total weapon destiny (before any draws, chance before each draw, or after all draws).

billbisco:
@BOTVHD  @Gergall   If you want me to remove the updated "Errata" Text i can from this PR.  Not a big deal.  I simply saw the mismatch between the old text and the "Errata" text and wanted to fix it as well.  If you want to update / fix in a later update, it's also fine.  

Regarding timing, I have always assumed that unless a card specifically says that it may be used to affect the result even after the destiny draw, that modifiers are always utilized before the actual draw result.  This being said, I don't think there's any utility to deciding this later.  If you want to use Stay Sharp! and miss then you simply would choose not to add the +2 bonus at all.  If you want to use Stay Sharp! and hit then you want to utilize the +2 bonus.  I cannot think of a situation where, even if it was rules supported, that you would want to see the draw and then decide to add the bonus or not.  

I think the current implementation is correct, but I am of course all ears to other arguments.

BOTVHD:
> I cannot think of a situation where, even if it was rules supported, that you would want to see the draw and then decide to add the bonus or not.

There are definitely weapons where you want to have a specific value (ex: Concussion Grenade, Energy Shell Launchers, etc) and would prefer to make the choice as late as possible.

billbisco:
Oooh  Good Catch @BOTVHD .   Well, the above examples don't work for starships.  There are no LS Starship weapons where having a destiny other than above targets Defensive Value would do something useful.  The only exception I can see the Dark Side Card:  Proton Bombs which Light Side could technically steal via Cloud City Sabacc.  However, Bombing Run requires the Dark Side Effect and I don't think there's any current way to steal that DS effect for LS.  

I still think that text to apply modifiers at a weird time has to be specific to the card: ( Example:  A Few Maneuvers or Don't Tread on Me).  Because Stay Sharp!'s wording is not like Don't Tread on Me, I still think the modifier has to be applied before Weapon Destiny is Drawn.

BOTVHD:
> Well, the above examples don't work for starships. There are no LS Starship weapons where having a destiny other than above targets Defensive Value would do something useful.

Yes, but... even if there are no cards that currently exploit/showcase a problem, we still want to "get it right" for several reasons:
1. A virtual card could be made where it becomes relevant (some starship weapon: "if destiny = defense value, something happens").
2. This card could be referenced or used as a template for making a card that interacts with non-starship weapons.
3. Consistent/correct implementation to the rules (even when the outcome/result doesn't matter) may give a higher chance of any AI generated code (based on its analysis of the code base) being correct.

> I still think that text to apply modifiers at a weird time has to be specific to the card: ( Example: A Few Maneuvers or Don't Treat on Me). Because Stay Sharp!'s wording is not like Don't Tread on Me, I still think the modifier has to be applied before Weapon Destiny is Drawn.

I think that's a reasonable path forward.

PR #1035, opened 2026-09-10
Implement Yaggle Gakkle (3_142)

### LINKED ISSUE #97: Yaggle Gakkle

billbisco:
## Summary
Implements Hoth Dark Used Interrupt **Yaggle Gakkle** (GEMP `3_142`) and owns the shared creature **get-ferocity** extract used by Attack, Yaggle, and Disarming Creature.

**Closes #97**

### Card text
Destiny 6 Dark Used Interrupt (R2). Target a creature vehicle at same site as a creature. If ferocity > target's maneuver + landspeed, creature vehicle is eaten, cumulatively adding 2 to creature's ferocity.

### What changed
- **Shared ferocity path**
  - `DrawFerocityDestinyEffect` — draws printed ferocity destinies (if any), then Disarming-style `SUBTRACT_DESTINY_FROM_FEROCITY` draws
  - `CalculateFerocityEffect` — wraps that destiny path + modifiers into a reusable final ferocity callback (`ferocityCalculated`)
  - Attack continues to use `DrawFerocityDestinyEffect`; Yaggle calls `CalculateFerocityEffect` (no card-local ferocity destiny draws)
- **Yaggle Gakkle (`Card3_142`)** — eat threshold via shared get-ferocity; cumulative +2 via card-local `FerocityModifier`
- **Disarming Creature (`Card3_033`)** — Immediate Effect attach; subtract destiny on each ferocity calculation through the same path
- No `MAY_MOVE_OTHER_CARD_AS_REACT_FROM_ATTACK` / second attack-react path

### Tests
`Card_3_142_Tests` — **26/0**
`Card_3_033_Tests` — **20/0** (shared-path verification)

VHD / doc cases covered:
- Same-site play restriction; self/opponent creature and vehicle targeting; multiple choices
- Reject creature vehicle without maneuver (Ronto)
- Threshold equal fail vs greater-than eat; cumulative +2 / multiple copies
- Ferocity bonus ends when creature is lost; rider jump-off; Sense cancel; Revert
- Disarming subtraction on shared CalculateFerocity path (prevent eat / still allow eat + stack +2)
- Zero/negative effective ferocity fails threshold
- Attack after eat uses cumulative +2
- **VHD:** Bubo→Eopie (no ferocity destiny draws); Krayt→Eopie (2 draws); Krayt+Disarming→Eopie with cumulative prior Yaggle (2 positive + 1 negative + stacked +2)
- Stats / BlueprintIconCheck + BlueprintKeywordCheck

### Skips / N/A
- LS fixed-ferocity creature **success** eat: Worrt (3) only equals Bantha threshold (3); covered as equal-fail ownership path instead
- No `@Ignore` on card-text cases

### Confirm
- No second attack-react path invented
- Yaggle does not invent card-local ferocity destiny draws


BOTVHD:
Hmm, I'm not sure why this is doing a ferocity destiny draw - it should just be using the ferocity of the creature (which sometimes involves drawing ferocity destinies, but many times does not).  The card itself doesn't directly cause any destiny draws.

I think a bunch of back end stuff needs to be reworked to break the ferocity calculation (including destinies and modifiers) out of the Attack logic and make it accessible to cards like this (as well as for use in the Attack section).  That 'get ferocity' thing will also handle any negative ferocity modifiers (like from Disarming Creature).

You can imagine a few cases like:
- Yaggle Gakkle using Bubo to target an Eopie (no ferocity destiny draws)
- Yaggle Gakkle using Krayt Dragon to target an Eopie (2 ferocity destiny draws)
- Yaggle Gakkle using Krayt Dragon with Disarming Creature to target an Eopie (2 positive ferocity destiny draws, 1 negative ferocity destiny draw, and cumulative modifiers from previous Yaggle Gakkle plays)

billbisco:
Aligned the stacked `Card3_033` deploy filter with https://github.com/PlayersCommittee/gemp-swccg-public/pull/1032.

This branch had `sameSiteAs` only; updated to `presentAt(sameSiteAs(...))` to match the dedicated PR.

PR #1038, opened 2026-09-10
4_178 IG-88's Pulse Cannon

### LINKED ISSUE #115: IG-88's Pulse Cannon

billbisco:
## Summary
Implements **IG-88's Pulse Cannon (4_178)**, Dagobah Dark Character Weapon. Closes #115.

## Game text
Use 1 Force to deploy on IG-88, 4 on your other warrior. Adds 2 to power. May target X non-droid characters or creatures using X Force. Draw destiny for each. If destiny = 0, character is power -1 and forfeit -1 until end of turn. If destiny -1 > defense value, target hit.

## What changed
- New weapon `Card4_178`.
- Card-scoped fire builder: `buildFireWeaponIG88sPulseCannonAction` / `IG88sPulseCannonDrawDestinyForTargets` (reuses existing `_numTargets`; not a broad multi-target engine rewrite).
- Free firing not allowed (X Force required).

## Tests (`Card_4_178_Tests`, 11 tests)
1. Stats/icons/keywords (VHD Blueprint checks)
2. Deploy cost 1 on IG-88
3. Deploy cost 4 on other warrior
4. Adds 2 to power while attached
5. Fire at X targets; pay X Force; select X distinct non-droid characters/creatures
6. Cannot target droids/vehicles/starships
7. Destiny per target; destiny 0 applies power/forfeit -1 (characters only)
8. destiny -1 > defense value hits; equal does not
9–11. Free fire / Defensive Fire (V) cannot use this weapon; other Mouse-style edges from Doc

## Known gaps
None required for first PR beyond Doc soft edges not yet expressed.

### DIFF
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set4/dark/Card4_178.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set4/dark/Card4_178.java
@@ -1,0 +1,92 @@
+package com.gempukku.swccgo.cards.set4.dark;
+
+import com.gempukku.swccgo.cards.AbstractCharacterWeapon;
+import com.gempukku.swccgo.cards.GameConditions;
+import com.gempukku.swccgo.common.ExpansionSet;
+import com.gempukku.swccgo.common.Icon;
+import com.gempukku.swccgo.common.Keyword;
+import com.gempukku.swccgo.common.PlayCardOptionId;
+import com.gempukku.swccgo.common.Rarity;
+import com.gempukku.swccgo.common.Side;
+import com.gempukku.swccgo.common.TargetingReason;
+import com.gempukku.swccgo.common.Uniqueness;
+import com.gempukku.swccgo.filters.Filter;
+import com.gempukku.swccgo.filters.Filters;
+import com.gempukku.swccgo.game.PhysicalCard;
+import com.gempukku.swccgo.game.SwccgGame;
+import com.gempukku.swccgo.logic.actions.FireWeaponAction;
+import com.gempukku.swccgo.logic.actions.FireWeaponActionBuilder;
+import com.gempukku.swccgo.logic.modifiers.DefinedByGameTextDeployCostModifier;
+import com.gempukku.swccgo.logic.modifiers.DefinedByGameTextDeployCostToTargetModifier;
+import com.gempukku.swccgo.logic.modifiers.Modifier;
+import com.gempukku.swccgo.logic.modifiers.PowerModifier;
+
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * Set: Dagobah
+ * Type: Weapon
+ * Subtype: Character
+ * Title: IG-88's Pulse Cannon
+ */
+public class Card4_178 extends AbstractCharacterWeapon {
+    public Card4_178() {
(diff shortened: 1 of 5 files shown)


BOTVHD:
Hmm although many weapons can target creatures, I thought creatures did not participate in battles (and would not be eligible targets during battle).  Aren't they only eligible targets during Sniper, Attacks, etc?

> All characters, starships, vehicles, weapons, or 
> devices at the location where a battle takes place 
> (whether present or not) are participating in that 
> battle (unless they are excluded)

I don't think creatures participate?

billbisco:
VHD addressed: creatures are excluded from battle fire targets (attack/sniper-only). Test `IG88sPulseCannonCannotTargetCreaturesDuringBattle` asserts wampa illegal in battle. SHA 5891fb5a028e1bb9ab6824e66d6d66b49b1eadfc — Card_4_178_Tests 11/11 green.

BOTVHD:
> VHD addressed: creatures are excluded from battle fire targets (attack/sniper-only). Test `IG88sPulseCannonCannotTargetCreaturesDuringBattle` asserts wampa illegal in battle. SHA [5891fb5](https://github.com/PlayersCommittee/gemp-swccg-public/commit/5891fb5a028e1bb9ab6824e66d6d66b49b1eadfc) — Card_4_178_Tests 11/11 green.

If it worked here, then it probably is broken for all the rest of the weapons that can target creatures.  If so, it should be fixed at a battle engine level and not using a card-specific workaround in the Pulse Cannon.

I'll try to follow up for rules/tests.  My thinking is that only cards participating in a battle can be targeted by weapons during a battle - and since creatures are not included in battles (I think!), they shouldn't be targetable by any weapon (even ones that allow targeting creatures) during a battle.

billbisco:
@BOTVHD    I think it would be worth looking into Firing at adjacent sites because in those cases, those characters aren't technically participating either but can still be shot at

billbisco:
@BOTVHD We consulted the 2023 Advanced Rulebook on this.

Creatures do not participate in battles, so weapons cannot normally target a creature during a battle. They participate in **attacks**. During an attack, a weapon may target a creature only if it says it targets creatures, and only participating cards may fire.

IG-88's Pulse Cannon's printed text does allow targeting creatures, so the card-local filter is:

- during battle: non-droid characters only
- not during battle (attack / sniper-style fire): also creatures

That matches the ARB. We do **not** want an engine-wide change that lets ordinary weapons target creatures in battle. If you still want a shared helper for "weapons whose text includes creatures," we can extract that, but we should not change targeting for weapons that do not mention creatures.

Firing at characters outside of battle still requires specific card text (Sniper, Weequay Marksman, etc.). Pulse Cannon does not grant that.

BOTVHD:
Most of the weapons out there that can target creatures just OR all the valid targets together like this for the weapon action builder:
`Filters.or(Filters.character, targetedAsCharacter, Filters.creature, Filters.vehicle)`

It just seems weird that IG's weapon has to specifically allow creature targeting only during ~~battle~~ non-battles...

billbisco:
@BOTVHD On the remaining targeting questions:

Creatures still do not participate in battles. Pulse Cannon only adds creatures to the fire filter when it is not during battle (attacks / sniper-style). That matches the 2023 ARB. We are not changing ordinary weapons to target creatures in battle.

Firing at characters at adjacent sites still needs specific card text (Sniper, Weequay Marksman, etc.). Pulse Cannon does not grant that.

If you want a shared helper for "weapons whose printed text includes creatures," we can extract Pulse Cannon''s battle-vs-attack filter. We should not OR creatures into every weapon''s battle targets.

PR #1043, opened 2026-09-11
Add This Is No Cave (4_63)

### LINKED ISSUE #108: This Is No Cave

billbisco:
﻿## Summary
Implements **This Is No Cave** (Dagobah Light Used Interrupt, blueprint `4_63`).

- Relocate one starfighter from a Big One to the related Asteroid Cave / Space Slug Belly (or vice versa) for free.
- Opens the related Space Slug's mouth if it was closed.
- If a smuggler is aboard, the starfighter is immune to attrition and subtracts 5 from asteroid destiny targeting that starfighter until end of your next turn.
- OR cancel Corrosive Damage (already implemented on master).

Closes #108.

## Test plan
- [ ] Blueprint/icon/keyword checks for Used Interrupt destiny 5
- [ ] Relocate Big One → related belly and belly → related Big One
- [ ] Not playable without related site / wrong location / capital (Medium Transport)
- [ ] Opens closed mouth; leaves open mouth open
- [ ] Related-only with multiple Big Ones on table
- [ ] Smuggler aboard buffs duration through end of your next turn; no smuggler → no buffs
- [ ] Cancel Corrosive Damage on table; not playable as cancel when absent

### DIFF
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set4/light/Card4_063.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set4/light/Card4_063.java
@@ -1,0 +1,196 @@
+package com.gempukku.swccgo.cards.set4.light;
+
+import com.gempukku.swccgo.cards.AbstractUsedInterrupt;
+import com.gempukku.swccgo.cards.GameConditions;
+import com.gempukku.swccgo.common.ExpansionSet;
+import com.gempukku.swccgo.common.Icon;
+import com.gempukku.swccgo.common.Rarity;
+import com.gempukku.swccgo.common.Side;
+import com.gempukku.swccgo.common.Title;
+import com.gempukku.swccgo.common.Uniqueness;
+import com.gempukku.swccgo.filters.Filter;
+import com.gempukku.swccgo.filters.Filters;
+import com.gempukku.swccgo.game.PhysicalCard;
+import com.gempukku.swccgo.game.SwccgGame;
+import com.gempukku.swccgo.logic.GameUtils;
+import com.gempukku.swccgo.logic.TriggerConditions;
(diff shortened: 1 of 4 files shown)


BOTVHD:
Per this rules clarification https://forum.starwarsccg.org/viewtopic.php?t=86890

> TIEs cannot exist at the Asteroid Cave/Belly and so are not valid targets to try to relocate there

This should be a pretty easy test to add.  I think the existing test checks that you can target a TIE to relocate from belly/cave to sector (which is good), but the opposite (relocate TIE from sector to belly/cave) should not work.  Hopefully, the existing 'can relocate' checks that This Is No Cave uses are already handling this correctly.

billbisco:
Thank you, VHD. TIEs cannot exist at Asteroid Cave / Space Slug Belly, so they are not valid This Is No Cave targets toward that site.

The existing can-relocate check was still offering a TIE at Big One. Relocate targeting now rejects TIEs when the destination is the cave/belly. TIE from belly to Big One still works.

BOTVHD:
> Thank you, VHD. TIEs cannot exist at Asteroid Cave / Space Slug Belly, so they are not valid This Is No Cave targets toward that site.
> 
> The existing can-relocate check was still offering a TIE at Big One. Relocate targeting now rejects TIEs when the destination is the cave/belly. TIE from belly to Big One still works.

Although this fix works, it probably should be fixed at a higher level (up the 'can relocate to' chain so that all cards checking to see if a TIE can be relocated to a specified location will correctly report 'no').  Maybe should be handled in getValidMoveTargetFilterForCardType for AbstractStarship, but I'm not certain about that, yet.  The landing prevention on TIEs seems to be defined in tieAllowedToLand() and mayNotLandFromLocationToLocation().

For now, the workaround in This Is No Cave seems OK until the upstream architecture fixes this.  Recommend no architecture change, yet, until we identify exactly where/how this should be handled.

PR #1044, opened 2026-09-11
5_137 Double-Crossing, No-Good Swindler (Closes #126)

### LINKED ISSUE #126: Double-Crossing, No-Good Swindler

billbisco:
Why: leftover #126. Double-Crossing, No-Good Swindler was unimplemented. After Nabrun transports, you may immediately deploy cards to that site from hand; troopers deploy free. Disarmed is allowed outside Control. Deploying on a location (Presence Of The Force) is not deploying to that site.

What it Does: Adds Card5_137. Action 1: Han and your Lando at same site, opponent loses 3 Force. Action 2: Nabrun just completed a transport, Nabrun is lost, then optional repeating deploy from hand to that site. DeployCardsToLocationFromHandEffect offers only cards that can deploy to that site or onto a card there. Adds Card_5_137_Tests, including SwindlerDisarmedDeploysOutsideControlPhase and SwindlerPresenceOfTheForceNotOfferedBecauseOnLocationIsNotToLocation.

What it Fixes: Swindler can be played. Disarmed may deploy via the Nabrun action outside Control. Presence Of The Force and Den Of Thieves are not offered for that site deploy.

Issues: Closes #126

### DIFF
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set5/dark/Card5_137.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set5/dark/Card5_137.java
@@ -99,9 +99,14 @@ public class Card5_137 extends AbstractLostInterrupt {
                                     new AddUntilEndOfCardPlayedModifierEffect(action, self,
                                             new DeploysFreeModifier(self, Filters.and(Filters.your(self), Filters.trooper)),
                                             "Troopers deploy free"));
-                            // May repeatedly deploy from hand to destination site (not as a react)
+                            // May repeatedly deploy from hand to destination site (not as a react).
+                            // Limit to cards that can actually deploy to a site / cards there — excludes
+                            // side-of-table Effects (e.g. Den Of Thieves).
+                            Filter deployableToSite = Filters.or(Filters.character, Filters.starship, Filters.vehicle,
+                                    Filters.weapon, Filters.device, Filters.and(Filters.Effect,
+                                            Filters.or(Filters.deploys_on_location, Filters.deploys_on_characters)));
                             action.appendEffect(
-                                    new DeployCardsToLocationFromHandEffect(action, playerId, Filters.any, thatSite, false));
+                                    new DeployCardsToLocationFromHandEffect(action, playerId, deployableToSite, thatSite, false));
                         }
                     }
             );
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set5/dark/Card5_137.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set5/dark/Card5_137.java
@@ -1,0 +1,112 @@
+package com.gempukku.swccgo.cards.set5.dark;
+
+import com.gempukku.swccgo.cards.AbstractLostInterrupt;
+import com.gempukku.swccgo.cards.GameConditions;
+import com.gempukku.swccgo.common.ExpansionSet;
+import com.gempukku.swccgo.common.Icon;
+import com.gempukku.swccgo.common.Rarity;
+import com.gempukku.swccgo.common.Side;
+import com.gempukku.swccgo.common.Title;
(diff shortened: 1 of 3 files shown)


BOTVHD:
Probably needs some tests exercising the deploy from hand part.  Things that deploy as if from hand (like Den Of Thieves) should also work.  Disarmed should be deployable (even though not during Control Phase!), POTF should not, simultaneous deployment should work (although this is not very common to sites).

src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set5/dark/Card5_137.java
```diff
@@ -99,9 +99,14 @@ protected void performActionResults(Action targetingAction) {
                                     new AddUntilEndOfCardPlayedModifierEffect(action, self,
                                             new DeploysFreeModifier(self, Filters.and(Filters.your(self), Filters.trooper)),
                                             "Troopers deploy free"));
-                            // May repeatedly deploy from hand to destination site (not as a react)
+                            // May repeatedly deploy from hand to destination site (not as a react).
+                            // Limit to cards that can actually deploy to a site / cards there — excludes
+                            // side-of-table Effects (e.g. Den Of Thieves).
+                            Filter deployableToSite = Filters.or(Filters.character, Filters.starship, Filters.vehicle,
```
BOTVHD, line 105: Instead of trying to filter here, shouldn't there be checks in DeployCardsToLocationFromHandEffect that limit the cardFilter to only cards that actually deploy to the location?

src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set5/dark/Card_5_137_Tests.java
```diff
@@ -257,4 +260,245 @@ public void DoubleCrossingPlayableAfterNabrunTransportNabrunLost() {
...
+
+		scn.SkipToLSTurn(Phase.MOVE); // outside Control
+		assertTrue("Sanity: Nabrun during Move (outside Control)", scn.GetCurrentPhase() == Phase.MOVE);
+		assertTrue(PlaySwindlerAfterNabrunToDeployWindow(scn));
+
+		boolean sawDeploy = false;
+		for (int j = 0; j < 20; j++) {
+			if (AtDeployFromHandChoice(scn) || scn.DSDecisionAvailable("deploy") || scn.DSHasCardChoiceAvailable(trooper2)) {
+				sawDeploy = true;
+				assertFalse("Disarmed must not deploy via Swindler outside Control phase",
```
BOTVHD, line 380: This is wrong, according to: https://forum.starwarsccg.org/viewtopic.php?p=1450011#p1450011

> 2) Yes, Disarmed is fine - not limited to control phase. Most cards normally deploy during the deploy phase - DCNGS clearly circumvents regular deployment timing rules.

src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set5/dark/Card5_137.java
billbisco: Moved the site-capability check into `DeployCardsToLocationFromHandEffect`. It ANDs `deployableToLocation` and keeps Effects that deploy on a card at the site (Disarmed). Side-of-table Effects and Effects that deploy on the location itself are not offered. Thanks VHD.

src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set5/dark/Card_5_137_Tests.java
billbisco: Agreed. Printed Disarmed is control-phase, but this interrupt says immediately deploy to that site, so it ignores the card's normal deploy-phase timing. `SwindlerDisarmedDeploysOutsideControlPhase` now offers Disarmed during Move when both players have a character with a weapon at the destination. Presence Of The Force is not offered: deploying on a location is not deploying to that site. Thanks VHD.

PR #1047, opened 2026-09-11
5_52 Impressive, Most Impressive

### LINKED ISSUE #120: Impressive, Most Impressive

billbisco:
## Summary
Implements **Impressive, Most Impressive (5_52)**, Cloud City Light Lost Interrupt. Closes #120.

## Game text
If opponent just attempted to freeze a character using Carbon-Freezing or All Too Easy, draw destiny; add character ability; if total > 7, cancel effect, release if captive, and the character **'hides' (may not participate in battle) for remainder of turn**.

## What changed
- New interrupt `Card5_052`.
- Opens Carbon-Freezing attempt response window by switching `Card5_114` from `UnrespondableEffect` to `RespondableEffect` (targeting already performed; required for interrupt responses).
- **Chief (C):** replaces `MayNotBattleUntilEndOfTurnEffect` with new shared **`HideUntilEndOfTurnEffect`** in the same AboutToHide / hide family as `HideFromBattleEffect` (emits `AboutToHideFromBattleResult`, then applies may-not-participate until end of turn; if currently in battle, also excludes from that battle).
- Narrow hook: `TriggerConditions.isAboutToHideFromBattle` no longer requires the card to already be participating in the current battle, so **Nice Of You Guys To Drop By** can cancel out-of-battle hide-until-EOT (e.g. Impressive during Carbon-Freezing control). `Card3_046` itself unchanged — it already listens to AboutToHide.
- No seeker-ignore path (coordinate with Attark/Velken seeker-ignore work if related).

### Shared engine
- New: `HideUntilEndOfTurnEffect` (AboutToHide-cancelable hide lasting remainder of TURN — not battle-only `HideFromBattleEffect`).
- `TriggerConditions.isAboutToHideFromBattle` broadened slightly for hide-until-EOT outside an active battle.
- Narrow Carbon-Freezing response enablement on `Card5_114`. Reviewers: other cards that respond to Carbon-Freezing / All Too Easy freezes depend on that window.

## Tests (`Card_5_052_Tests`, 6 tests)
1. Stats / keywords
2. Not playable without freeze attempt
3. Success: destiny+ability >7 cancels Carbon-Freezing, releases captive, hide-until-EOT applied
4. Fail ≤7 leaves captive
5. Hide-until-EOT clears after the turn ends
6. Nice Of You Guys To Drop By cancels Impressive's AboutToHide hide-until-EOT

Maven: `Card_5_052_Tests` — Tests run: 6, Failures: 0, Errors: 0.

## Known gaps
All Too Easy success path not separately asserted in suite (CF path covers cancel/release/hide). Doc soft edges may still be open.

### DIFF
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set5/dark/Card5_114.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set5/dark/Card5_114.java
@@ -31,7 +31,7 @@ import com.gempukku.swccgo.logic.effects.FreezeCharacterEffect;
 import com.gempukku.swccgo.logic.effects.LoseCardFromTableEffect;
 import com.gempukku.swccgo.logic.effects.LoseForceEffect;
 import com.gempukku.swccgo.logic.effects.TargetCardOnTableEffect;
-import com.gempukku.swccgo.logic.effects.UnrespondableEffect;
+import com.gempukku.swccgo.logic.effects.RespondableEffect;
 import com.gempukku.swccgo.logic.timing.Action;
 import com.gempukku.swccgo.logic.timing.EffectResult;
 import com.gempukku.swccgo.logic.timing.GuiUtils;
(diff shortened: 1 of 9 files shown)


BOTVHD:
Probably needs some rules clarifications about whether this is the same 'hide' mechanic as others (optional top level actions during battle) and should be prevented with Nice Of You Guys To Drop By, for example.  This card's 'hide' mechanic feels a little different because it is applied as a modifier before a battle even starts (rather than an action taken during battle).

If so, then I would expect this card should apply a 'hides' effect rather than "a may not participate in battle" - and then cards that have the hiding modifier can be handled as not participating in battle?

PR #1048, opened 2026-09-11
5_23 Frozen Assets (Closes #118)

### LINKED ISSUE #118: Frozen Assets

billbisco:
## Summary
Implements Cloud City Effect **Frozen Assets** (5_23). Closes #118.

### Table mode
- Deploy on your side of table
- At sites with a frozen captive: your Rebels deploy -2 and are power +2 in battle

### Force-pile mode (VHD)
- Deploy freezes opponent Force into `FROZEN_PILE` (order preserved)
- **Frozen Assets sits Active at TOP of `FROZEN_PILE`** (not Force Pile) so FA is out of Force Pile shuffle/draw/use paths
- Usable Force = Force Pile only; frozen Force may be lost as life force only when usable Force Pile is empty; frozen Force still counts toward life force (FA marker does not)
- Empty Force Pile deploy OK (FA alone on Frozen Pile)
- Slip Sliding Away relocates FA when it is conceptual top (Force Pile empty) -> unfreezes; FA ends at bottom of Force Pile
- **Beggar may NOT use Frozen Force** (AR universal)
- UI sandwich: Frz (frozen Force count) + usable Force + total

## Tests
`Card_5_23_Tests` - 13/0/0

## Notes
Independent PR vs master. Overlay 17003.


BOTVHD:
I think I see that Frozen Assets is ending being tracked as the bottom card of the Force Pile (and so special workarounds are needed to keep it from getting shuffled, etc).  It may be better to make Frozen Assets the top card of the Frozen Pile instead, so that it won't be involved in any Force Pile interactions.  Hopefully, that means less manipulation/workaround for the core mechanics and is easier to read/maintain/etc and fix any issues specific to Frozen Assets.

BOTVHD:
"Force below this card may not be drawn or used." sounds like it applies to universally (and the AR entry for Frozen Assets reinforces this) so Beggar should not be allowed to use any Frozen Force.  The workaround exceptions to let LS use the opponent's Frozen Force don't make sense to me.  Is there a ruling allowing that?

I see the comment:

> // Beggar AR: may use both usable Force Pile and Frozen Pile

But after looking at the AR for both Beggar and Frozen Assets, I don't see any mention of this...

billbisco:
**@Test restore tip** `08963ba` — `Card_5_23_Tests` **13/0/0**. Restored `@Test` on `ForcePileModeFaStaysOnFrozenTopNotInForcePile` + sandwich UI stats; Beggar VHD case now attaches Beggar on Marketplace so usable-only pool assert is real.

src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/game/state/GameState.java
```diff
@@ -2905,10 +2924,46 @@ public int getForcePileSize(String playerId) {
...
+        return getCardPileSize(playerId, Zone.FROZEN_PILE);
+    }
+
+    /**
+     * Frozen Force only — excludes Frozen Assets marker sitting at top of FROZEN_PILE.
+     */
+    public int getFrozenForceCount(String playerId) {
+        int count = 0;
+        for (PhysicalCard c : getFrozenPile(playerId)) {
+            if (!"Frozen Assets".equals(c.getTitle()))
```
BOTVHD, line 2937: Should use `Title.Frozen_Assets` instead (if planning on specifying a title).  Frozen Assets almost behaves like an insert (a face up card in opponent's pile that doesn't count toward the pile total) and there may be good ways to filter it out of counts like an insert?  I haven't looked closely at that, though.

src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/game/state/GameState.java
```diff
@@ -2905,10 +2924,46 @@ public int getForcePileSize(String playerId) {
...
+                return c;
+        }
+        return null;
+    }
+
     public PhysicalCard getBottomOfCardPile(String playerId, Zone zone) {
         List<? extends PhysicalCard> cards = getCardPile(playerId, zone);
         if (cards==null || cards.isEmpty())
             return null;
+        // Tikkes: bottom-most UNFROZEN (usable) Force card — skip Frozen Assets marker
```
BOTVHD, line 2958: This workaround shouldn't be needed anymore, with FA in the frozen pile, right?

src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/game/state/GameState.java
```diff
@@ -4596,8 +4665,15 @@ public void playerActivatesForce(String player, boolean firstActivated, boolean
 
     public void playerUsesForce(String player, boolean firstUsed, boolean lastUsed) {
         List<PhysicalCard> forcePile = _forcePiles.get(player);
-        if (!forcePile.isEmpty()) {
-            PhysicalCard card = forcePile.get(0);
+        // Skip Frozen Assets Effect sitting on Force Pile (not usable as Force)
```
BOTVHD, line 4668: Again, like other areas, this workaround can be reverted back to normal, now that FA exists in the Frozen Pile (not Force Pile), right?

BOTVHD:
> I think I see that Frozen Assets is ending being tracked as the bottom card of the Force Pile (and so special workarounds are needed to keep it from getting shuffled, etc). It may be better to make Frozen Assets the top card of the Frozen Pile instead, so that it won't be involved in any Force Pile interactions. Hopefully, that means less manipulation/workaround for the core mechanics and is easier to read/maintain/etc and fix any issues specific to Frozen Assets.

I think there are still a bunch of workarounds that assume FA is going to be at the bottom of the Force Pile.  With FA at the top of the Frozen Pile instead, those can be reverted back to normal.

Also, can you post a small crop of what the UI currently looks like for the 5 different cases?
- No Frozen Force, No Force Pile
- No Frozen Force, Force Pile with X cards
- Frozen Force with Y cards + FA on top, No Force Pile
- Frozen Force with Y cards + FA on top, Force Pile with X cards
- Frozen Force with 0 cards + FA on top, Force Pile with X cards

billbisco:
Thank you, VHD.

Title.Frozen_Assets is now used instead of a title string.

The remaining Force Pile skips stay for Slip Sliding Away: that relocate puts Frozen Assets at the bottom of Force Pile after unfreeze, so Force Pile still has to treat the Effect as not Force (Tikkes bottom card, life Force, use, shuffle). While frozen, Frozen Assets stays on Frozen Pile, not Force Pile.

Sandwich UI now also lights when Frozen Assets is the only card on Frozen Pile (0 frozen Force). The five cases:
- No Frozen Assets, empty Force Pile: normal 0, no sliver
- No Frozen Assets, Force Pile X: normal X, no sliver
- Frozen Y + Frozen Assets, empty Force Pile: Frz Y | Use 0 (Y), sliver
- Frozen Y + Frozen Assets, Force Pile X: Frz Y | Use X (X+Y), sliver
- Frozen 0 + Frozen Assets, Force Pile X: Frz 0 | Use X (X), sliver

I do not have a live table crop of Frozen Assets to post here yet.

BOTVHD:
> Thank you, VHD.
> 
> Title.Frozen_Assets is now used instead of a title string.
> 
> The remaining Force Pile skips stay for Slip Sliding Away: that relocate puts Frozen Assets at the bottom of Force Pile after unfreeze, so Force Pile still has to treat the Effect as not Force (Tikkes bottom card, life Force, use, shuffle). While frozen, Frozen Assets stays on Frozen Pile, not Force Pile.

Instead of Slip Sliding Away sending Frozen Assets to the bottom of the Force Pile, I think it would be better to keep it in the Frozen Pile.  Slip Sliding Away can move FA to the bottom of the Frozen Pile and something can automatically move the newly unfrozen force back to the Force Pile - or Slip Sliding Away can just directly unfreeze all the 'frozen force' cards in the Frozen Pile (leaving FA in the Frozen Pile) and then let FA self-clear at the end of the turn like normal.

A played Frozen Assets would never exist in the Force Pile, so no workarounds/skips would be needed.

That makes everything a lot simpler and more consistent (including for the UI).

billbisco:
Thank you, VHD. Slip Sliding Away now unfreezes by returning frozen Force to Force Pile and leaves Frozen Assets on Frozen Pile until end of turn. Frozen Assets never sits on Force Pile, so those Force Pile skips are gone.

src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/logic/effects/UseOneForceEffect.java
```diff
@@ -7,27 +7,34 @@
 import com.gempukku.swccgo.logic.timing.Action;
 import com.gempukku.swccgo.logic.timing.results.UseForceResult;
 
+
```
BOTVHD: All changes to this file can be reverted, right?  UseOneForce strictly interacts with the Force Pile and there should be no purpose in an 'allowFrozen' parameter...

This is probably leftover code from the Beggar workaround thing?

src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/logic/modifiers/querying/Force.java
BOTVHD: Should be able to revert all changes to this file?  Leftovers from workaround that incorrectly allowed Beggar to use frozen force?

src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/game/state/GameState.java
```diff
@@ -2905,6 +2924,44 @@ public int getForcePileSize(String playerId) {
...
+     * True when Frozen Assets is sitting as the marker at top of FROZEN_PILE.
+     */
+    public boolean hasFrozenAssetsMarker(String playerId) {
+        return isFrozenAssets(getTopOfFrozenPile(playerId));
+    }
+
+    /**
+     * Frozen Force only — excludes Frozen Assets marker sitting at top of FROZEN_PILE.
+     */
+    public int getFrozenForceCount(String playerId) {
```
BOTVHD: For consistency with other similar calls (getReserveDeckSize(playerId), _forcePiles.get(playerId).size(), etc) I think it may be better to name this getFrozenForceSize ?

getReserveDeckSize does something very similar where it returns the size (without including insert cards that may exist in the Reserve Deck)

BOTVHD:
I have an idea that may further simplify things quite a bit.

1. Change Slip Sliding Away to be able to target the top card of the Frozen Pile _if and only if_ the Force Pile is empty. This should allow the selection of Frozen Assets.  Then all workarounds for treating Frozen Assets as being in the Force Pile can be removed.

2. In Frozen Assets getGameTextRequiredAfterTriggers, in addition to the leaving table / cancel checks causing a moveFrozenPileToForcePile, include an additional check if self (FA) is not currently the top card of the Frozen Pile.  That way, when Slip Sliding Away sends FA to the bottom of the Frozen Pile, FA will auto-thaw the cards above.  This allows removing (or at least greatly simplifying) the chunk of logic in PlaceTopCardFromCardPileOnBottomOfCardPileEffect.java and keeps the 'thawing' calls within FA.

3. In ChooseExistingCardPileEffect, change the new includeFrozenAssetsAsForcePileTop boolean parameter to an includeFrozenPile parameter (still defaulting to false in the overloads).  Conditionally include Zone.FROZEN_PILE in the allowed filter (based on this parameter).  Remove all the workarounds (addConceptualFrozenAssetsForcePileTops, etc) that were needed to fake the existence of FA in the Force Pile.

BOTVHD:
Another step to take, eventually, is to scour the card database for anything that interacts with 'Force Pile' (or 'any' pile) and 'Lost' (or 'Lose') (example: Deneb Both, Plo Koon (V), etc) and make sure the card game text interacts with empty Force Pile correctly (routing to use Frozen Pile instead).

billbisco:
Thank you, VHD. Took the simplification as a whole.

Slip Sliding Away can choose Frozen Pile if and only if Force Pile is empty. Frozen Assets auto-thaws when it is not Frozen Pile top, and is lost at end of opponent's next turn. ChooseExistingCardPile uses includeFrozenPile (no fake Force Pile top). PlaceTopCard and UseForce/Force are generic again. Frozen count is getFrozenForceSize, same idea as getReserveDeckSize.

Sandwich UI (still no live table crop):
- No Frozen Assets, empty Force Pile: 0, no sliver
- No Frozen Assets, Force Pile X: X, no sliver
- Frozen Y + Frozen Assets, empty Force Pile: Frz Y | Use 0 (Y), sliver
- Frozen Y + Frozen Assets, Force Pile X: Frz Y | Use X (X+Y), sliver
- Frozen 0 + Frozen Assets, Force Pile X: Frz 0 | Use X (X), sliver

Empty Force Pile lose already routes to frozen Force. Scouring other Force Pile / lose cards (Deneb Both, Plo Koon (V), etc.) is later work.

PR #1049, opened 2026-09-11
Encode Attark + Velken Tezeri with shared Seeker ignore

### LINKED ISSUE #130: Attark

### LINKED ISSUE #139: Velken Tezeri

billbisco:
﻿## Summary
Encodes Attark (6_4 / #130) and Velken Tezeri (6_128 / #139) together, plus the Seeker updates they need.

One shared owner-scoped `SeekersMayIgnoreTargets` flag and a single all-or-nothing IgnoreList path on `AbstractSeeker` for Han / Leia / Luke / Motti / Tagge / Tarkin Seekers. No second ignore system.

When Attark or Velken is at a site you control, your Seekers may deploy free there and move for free. While Attark/Velken is active, your Seekers can ignore all current potential targets and later stop ignoring as a top-level action. If Attark/Velken leaves play or becomes inactive, the flag drops and ignoring stops.

## Test plan
- [x] `Card_6_004_Tests` (6) and `Card_6_128_Tests` (5) green on overlay 17004 / clean mount
- [x] Deploy-free / move-free / owner-scoped ignore flag / leave-table stops ignore
- [ ] Spot-check playable on local overlay 17004 if desired

Closes #130
Closes #139


BOTVHD:
Just looking through, I like this approach a lot - moves the shared mechanics of seekers into the AbstractSeeker area, which should make them more maintainable (hard to imagine any changes to a seeker card beyond the valid targeting that would not apply to all seekers).

There is an initial test to see if the ignore modifier turns off when the character leaves the table, but probably needs some additional tests about the seeker ignore list interactions (seeker moving to a different location, new characters being deployed/relocated, etc.).

I'm not sure if the OptionalGameTextTriggerAction (AbstractSeeker line 154) to ignore targets section is actually needed.  If any non-ignored target is present, there should be a required action anyway (either ignore all targets or choose a target to be lost) and the getGameTextRequiredAfterTriggers (AbstractSeeker line 175) should kick in, I think?

billbisco:
VHD: expanded ignore-list tests (retain while present, clear on leave/relocate, new target not covered). Optional ignore-all kept — required-after only forces make-lost and does not offer ignore-all.

PR #1050, opened 2026-09-11
6_46 Yarkora

### LINKED ISSUE #134: Yarkora

billbisco:
Why: leftover #134. Yarkora was unimplemented. Printed text: during your control phase, if at same site as an Undercover spy, may draw destiny; each of your Yarkoras on table may cumulatively subtract one from that destiny; cover is broken if destiny equals spy ability.

What it Does: Adds Card6_046. Control-phase action targets an opponent's undercover spy at same site. The acting Yarkora grants each of your Yarkoras on table an optional -1 on that specific destiny draw only (including a copy whose game text is canceled; not a nested Sense or later destiny). Cover breaks when destiny equals ability. Adds Card_6_046_Tests, including YarkoraCanceledCopyStillGetsGrantedSubtract and YarkoraSubtractIsNotOfferedOnALaterDestiny.

What it Fixes: Yarkora can break cover. Canceled copies still receive the granted subtract. The -1 does not apply to other destiny draws.

Issues: Closes #134

### DIFF
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set6/light/Card6_046.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set6/light/Card6_046.java
@@ -60,20 +60,21 @@ public class Card6_046 extends AbstractAlien {
      * Excludes inactive (e.g. missing) and supporting cards via Filters.filterActive.
      */
     static Collection<PhysicalCard> YarkoraGetActiveOwnedYarkoras(SwccgGame game, String playerId) {
-        return Filters.filterActive(game, null, Filters.and(Filters.owner(playerId), Filters.Yarkora));
+        return Filters.filterActive(game, null, Filters.and(Filters.owner(playerId), Filters.Yarkora, Filters.not(Filters.title("Yarkora"))));
     }
 
     /**
      * Action proxy granting each active owned Yarkora one optional -1 to this specific destiny draw.
-     * Species-based (not gametext), so Saelt-Marae / canceled gametext still eligible.
+     * Species-based for non-title-Yarkora (e.g. Saelt-Marae). Title Yarkora uses getGameTextOptionalAfterTriggers.
      */
     static ActionProxy YarkoraCreateSubtractActionProxy(final String playerId, final DrawDestinyState drawDestinyState) {
         return new AbstractActionProxy() {
             @Override
             public List<TriggerAction> getOptionalAfterTriggers(String playerId2, SwccgGame game, EffectResult effectResult) {
                 List<TriggerAction> actions = new LinkedList<TriggerAction>();
+                // Proxy only lives for this draw; match any non-canceled destiny drawn while active.
                 if (!playerId2.equals(playerId)
-                        || !TriggerConditions.isDestinyJustDrawn(game, effectResult, drawDestinyState)) {
+                        || !TriggerConditions.isDestinyJustDrawn(game, effectResult)) {
                     return actions;
                 }
                 for (PhysicalCard yarkora : YarkoraGetActiveOwnedYarkoras(game, playerId)) {
@@ -159,6 +160,21 @@ public class Card6_046 extends AbstractAlien {
                         }
                     }
             );
+            return Collections.singletonList(action);
+        }
+        return null;
+    }
+
+
+    @Override
+    protected List<OptionalGameTextTriggerAction> getGameTextOptionalAfterTriggers(final String playerId, SwccgGame game, EffectResult effectResult, final PhysicalCard self, int gameTextSourceCardId) {
+        // Each Card6_046 may -1 once per destiny drawn by your Yarkora (The One Rule / once per trigger).
+        if (TriggerConditions.isDestinyJustDrawnFor(game, effectResult, Filters.and(Filters.your(self), Filters.Yarkora))) {
+            final OptionalGameTextTriggerAction action = new OptionalGameTextTriggerAction(self, gameTextSourceCardId);
+            action.setText("Subtract 1 from destiny");
+            action.setActionMsg("Subtract 1 from destiny using " + GameUtils.getCardLink(self));
+            action.appendEffect(
+                    new ModifyDestinyEffect(action, -1, true));
             return Collections.singletonList(action);
         }
         return null;
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set6/light/Card6_046.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set6/light/Card6_046.java
@@ -1,0 +1,166 @@
+package com.gempukku.swccgo.cards.set6.light;
+
+import com.gempukku.swccgo.cards.AbstractAlien;
+import com.gempukku.swccgo.cards.GameConditions;
+import com.gempukku.swccgo.cards.effects.usage.OncePerPhaseEffect;
+import com.gempukku.swccgo.common.ExpansionSet;
+import com.gempukku.swccgo.common.Icon;
+import com.gempukku.swccgo.common.InactiveReason;
+import com.gempukku.swccgo.common.Phase;
+import com.gempukku.swccgo.common.Rarity;
+import com.gempukku.swccgo.common.Side;
+import com.gempukku.swccgo.common.Species;
+import com.gempukku.swccgo.common.SpotOverride;
+import com.gempukku.swccgo.common.Uniqueness;
+import com.gempukku.swccgo.filters.Filter;
+import com.gempukku.swccgo.filters.Filters;
+import com.gempukku.swccgo.game.AbstractActionProxy;
+import com.gempukku.swccgo.game.ActionProxy;
+import com.gempukku.swccgo.game.PhysicalCard;
+import com.gempukku.swccgo.game.SwccgGame;
+import com.gempukku.swccgo.game.state.DrawDestinyState;
+import com.gempukku.swccgo.game.state.GameState;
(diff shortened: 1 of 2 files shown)


src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set6/light/Card_6_046_Tests.java
```diff
@@ -0,0 +1,303 @@
...
+		scn.MoveCardsToLocation(site, yarkora);
+		MakeGarindanUndercoverAt(scn, site);
+
+		scn.SkipToLSTurn(Phase.DEPLOY);
+		assertFalse("Yarkora cover-break not available outside control phase",
+				scn.LSCardActionAvailable(yarkora, "Break a spy's cover"));
+	}
+
+	@Test
+	public void YarkoraCannotTargetOwnUndercoverSpy() {
```
BOTVHD, line 198: This test is basically empty - needs or put a LS spy undercover at the same site as Yarkora and then confirm no break cover is available to the Yarkora?

billbisco: Fixed — `YarkoraCannotTargetOwnUndercoverSpy` now puts LS Momaw Nadon undercover (LS Undercover `2_40`) at the same site and asserts Yarkora's break-cover action is unavailable (opponents-only filter).

src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set6/light/Card6_046.java
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * Active Yarkoras owned by the player that may optionally subtract from this destiny draw.
+     * Excludes inactive (e.g. missing) and supporting cards via Filters.filterActive.
+     */
+    static Collection<PhysicalCard> YarkoraGetActiveOwnedYarkoras(SwccgGame game, String playerId) {
+        return Filters.filterActive(game, null, Filters.and(Filters.owner(playerId), Filters.Yarkora, Filters.not(Filters.title("Yarkora"))));
+    }
+
+    /**
+     * Action proxy granting each active owned Yarkora one optional -1 to this specific destiny draw.
```
BOTVHD, line 67: This is a problem.  If a Yarkora A and Yarkora B are on table and Yarkora A takes the break cover action while Yarkora B's game text is canceled, Yarkora A should still grant the destiny modify optional action onto Yarkora B.

Also, if Yarkora (V) were created with different game text (lacking the destiny modifier text), a regular Yarkora (non-V) taking the action would still grant the Yarkora (V) the optional action...

src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set6/light/Card6_046.java
```diff
@@ -0,0 +1,182 @@
...
+     * Species-based for non-title-Yarkora (e.g. Saelt-Marae). Title Yarkora uses getGameTextOptionalAfterTriggers.
+     */
+    static ActionProxy YarkoraCreateSubtractActionProxy(final String playerId, final DrawDestinyState drawDestinyState) {
```
BOTVHD, line 70: Are there any ways this destiny modifier could affect more than the break undercover draw?  For example, if Light uses Yarkora to attempt to break a spies cover, Light draws destiny, Dark responds with an interrupt (based on the card drawn?), Light attempts to Sense the interrupt.  Are we sure Yarkora's destiny modifier proxy action isn't allowed to modify that Sense draw?

src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set6/light/Card6_046.java
billbisco: Agreed. Printed "Each of your Yarkoras on table may cumulatively subtract one from that destiny" is a grant from the Yarkora taking the action, not each copy using its own game text. The destiny-draw proxy now grants the optional -1 onto every one of your Yarkoras on table, including a copy whose game text is canceled. `YarkoraCanceledCopyStillGetsGrantedSubtract`. Thanks VHD.

src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set6/light/Card6_046.java
billbisco: The old proxy matched any destiny just drawn while it was registered, so a nested Sense draw during the cover-break destiny could have been offered the -1. It is now pinned to that DrawDestinyState (`isDestinyJustDrawn(..., drawDestinyState)`), so a later or nested destiny does not get it. `YarkoraSubtractIsNotOfferedOnALaterDestiny`. Thanks VHD.

BOTVHD:
Just double checking, but for a case with Yarkora A and Yarkora B at a site with opponent's undercover spy and Saelt-Marae on table:
- user initiates a break cover action on the the Yarkora A card
- user draws a destiny
- user has option of clicking on Yarkora A to take a 'subtract 1 from destiny' action
- user has option of clicking on Yarkora B to take a 'subtract 1 from destiny' action
- user has option of clicking on Saelt-Marae to take a 'subtract 1 from destiny' action
- user has option of passing after taking some (or none) of the optional subtraction actions
- user can only take the 'subtract 1 from destiny' action once on Saelt-Marae during this attempt

After completing this attempt and failing to break cover, Yarkora B can attempt to break cover and Yarkora A/Yarkora B/Saelt-Marae are each granted a an optional subtract 1 from destiny action (regardless of whether they used that action during Yarkora A's attempt earlier in the turn).

I think a test should be added to confirm that Yarkora can only take the optional, cumulative subtract action once during a break cover attempt.  To do this, have a 1 Yarkora on table draw destiny 3 against a spy's ability of 1 and attempt to subtract twice.  The second subtract should not be available and the destiny should be 2 (failed to break cover).

src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set6/light/Card_6_046_Tests.java
```diff
@@ -0,0 +1,409 @@
...
+		scn.LSChooseCard(garindan);
+
+		int taken = YarkoraTakeSubtractsAndFinishDestiny(scn, 2);
+		assertTrue("Expected two Yarkora subtract optionals, took " + taken, taken >= 2);
+		assertFalse("Two Yarkoras -1 each should make destiny 3 become 1 and break cover",
+				garindan.isUndercover());
+	}
+
+	@Test
+	public void YarkoraSaeltAlsoMaySubtract() {
```
BOTVHD: This is a bad test.  You could replace Saelt with anything (Han, for example) and this test would pass.  That is because the Yarkora on table (that is taking the break cover action) can provide a subtract 1 destiny action and perform the "2 - 1" destiny change regardless of whether Saelt is providing the option or not...

Change the prepared destiny to 3 and try to subtract twice (once from Yarkora and once from Saelt).  That change (along with adding a test to confirm that the subtract action is not offered more than once for each Yarkora during a draw action) should make this test actually work.

Perhaps a better way is to check the cards directly.  Start the break cover and action, pass until the destiny is just drawn, and then:
- assertTrue that Yarkora has a 'subtract 1' action available
- assertTrue that Saelt has a 'subtract 1' action available
- assertFalse that a rebelTrooper on table has a 'subtract 1' action available

billbisco:
Thank you, VHD. Added YarkoraMaySubtractOnlyOncePerBreakCoverAttempt: one Yarkora draws 3 against a spy of ability 1, subtracts once, cannot subtract a second time, and cover is not broken (destiny 2).

src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set6/light/Card_6_046_Tests.java
billbisco: Thank you, VHD. YarkoraSaeltAlsoMaySubtract now asserts subtract 1 on Yarkora and Saelt-Marae, and not on a Rebel Trooper. Destiny is 3; both subtracts are taken so cover breaks.

src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set6/light/Card6_046.java
```diff
@@ -0,0 +1,166 @@
...
+                                                        final PhysicalCard actingYarkora) {
+        return new AbstractActionProxy() {
+            @Override
+            public List<TriggerAction> getOptionalAfterTriggers(String playerId2, SwccgGame game, EffectResult effectResult) {
+                List<TriggerAction> actions = new LinkedList<TriggerAction>();
+                if (!playerId2.equals(playerId)
+                        || !TriggerConditions.isDestinyJustDrawn(game, effectResult, drawDestinyState)) {
+                    return actions;
+                }
+                for (PhysicalCard yarkora : YarkoraGetActiveOwnedYarkoras(game, playerId)) {
```
BOTVHD:
```suggestion
                for (PhysicalCard yarkora : Filters.filterActive(game, actingYarkora, Filters.and(Filters.owner(playerId), Filters.Yarkora))) {
```
and eliminate YarkoraGetActiveOwnedYarkoras?  I think this correctly identifies the actingYarkora as the source that is doing the active checking (instead of a 'null' for a game rule doing the checking, for example).

BOTVHD:
Yarkora is supposed to be a scout (according to scomplink) and needs the keyword added to the card and stats test.

PR #1061, opened 2026-09-11
Fix Panic (1_103) and Emergency Deployment (1_244) reveal/deploy/lose (Closes #303)

### LINKED ISSUE #303: Panic doesn't reveal the cards correctly.
Not positive of turn. DS Battle phase.
[https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=fungineer$cqx3id3l5y63gz5b](url)

Panic is used. The chat log displays "Draws card into hand" 3 times for panic, then I can deploy those where I want (if able). It should just reveal them and have me play from the pop-up. Unsure if opponent was unable to see the revealed cards or not (It sounded like no). 

*Check DS counter part too - emergency deployment*

billbisco:
Why: leftover #303. Panic and Emergency Deployment were not keeping revealed cards in Reserve, leftover undeployed cards were not always lost, and shuffle / leaving the pile did not end revealed-state. The helper name LoseTheRest was also wrong once leftover can stay on top.

What it Does: Shared reveal-deploy-leftover helper, renamed to DeployCardsFromRevealedReserveDeckEffect, with LeftoverMode LOSE vs LEAVE_ON_TOP. GameState marks revealed cards and clears them on shuffle or leaving the pile. Adds Card_1_103_Tests and Card_1_244_Tests, including PanicRevertAfterDeployThenDeclineLosesRemainingRevealed and PanicRevealedCardReturnedToReserveIsNotOfferedAgain.

What it Fixes: Revealed cards stay in Reserve in order. Shuffle ends the reveal even if order is unchanged. A revealed card that leaves Reserve (and even returns) is not offered again and is not leftover-lost. After revert, declining leftover deploys still loses remaining revealed cards.

Issues: Closes #303

### DIFF
--- a/src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/logic/effects/choose/DeployCardsFromReserveDeckAndLoseTheRestEffect.java
+++ b/src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/logic/effects/choose/DeployCardsFromReserveDeckAndLoseTheRestEffect.java
@@ -20,26 +20,45 @@ import java.util.LinkedList;
 import java.util.List;
 
 /**
- * Shared helper for Panic / Emergency Deployment / similar cards:
+ * Shared helper for Panic / Emergency Deployment / Go For Help / 3B3-21 / similar cards:
  * among a previously revealed set still in Reserve Deck, optionally deploy matching cards
- * (for free), then lose any remaining revealed cards still in Reserve Deck.
+ * (typically for free), then handle undeployed leftovers per {@link LeftoverMode}.
+ * <p>
+ * Revealed cards remain in Reserve Deck in the same order. If a response draws destiny (or
+ * otherwise moves a revealed card) mid-resolution, that card leaves the revealed set normally;
+ * remaining revealed cards stay deployable; leftovers are lost or left on top depending on mode.
+ * <p>
+ * Unpiloted ships/vehicles deploy via the normal play-card path, which may pair a pilot/driver
+ * from hand (normal cost) but not another revealed Reserve card.
  * <p>
  * Does not introduce persistent revealed-state engine tracking; uses existing reveal UI
  * plus deploy-from-reserve of specific cards (Sergeant Bruckman-style).
  */
 public class DeployCardsFromReserveDeckAndLoseTheRestEffect extends AbstractSubActionEffect {
+
+    /**
+     * What to do with revealed cards that remain undeployed (and still in Reserve Deck) at the end.
+     */
+    public enum LeftoverMode {
+        /** Lose remaining revealed cards still in Reserve Deck (Panic / Emergency Deployment / 3B3-21). */
+        LOSE,
+        /** Leave remaining revealed cards on top of Reserve Deck in the same order (Go For Help!). */
+        LEAVE_ON_TOP
+    }
+
     private final String _playerId;
     private final List<PhysicalCard> _remainingCards;
     private final Filterable _deployableTypesFilter;
     private final Filter _locationFilter;
     private final boolean _forFree;
+    private final LeftoverMode _leftoverMode;
 
     /**
      * Deploy matching revealed cards anywhere (for free by default for Panic-family), then lose the rest.
      */
     public DeployCardsFromReserveDeckAndLoseTheRestEffect(Action action, Collection<PhysicalCard> revealedCards,
                                                           Filterable deployableTypesFilter, boolean forFree) {
-        this(action, revealedCards, deployableTypesFilter, null, forFree);
+        this(action, revealedCards, deployableTypesFilter, null, forFree, LeftoverMode.LOSE);
     }
 
     /**
@@ -48,12 +67,35 @@ public class DeployCardsFromReserveDeckAndLoseTheRestEffect extends AbstractSubA
     public DeployCardsFromReserveDeckAndLoseTheRestEffect(Action action, Collection<PhysicalCard> revealedCards,
                                                           Filterable deployableTypesFilter, Filter locationFilter,
                                                           boolean forFree) {
+        this(action, revealedCards, deployableTypesFilter, locationFilter, forFree, LeftoverMode.LOSE);
+    }
+
+    /**
+     * Deploy matching revealed cards anywhere, then handle leftovers per leftoverMode.
+     */
+    public DeployCardsFromReserveDeckAndLoseTheRestEffect(Action action, Collection<PhysicalCard> revealedCards,
+                                                          Filterable deployableTypesFilter, boolean forFree,
+                                                          LeftoverMode leftoverMode) {
+        this(action, revealedCards, deployableTypesFilter, null, forFree, leftoverMode);
+    }
+
+    /**
+     * Deploy matching revealed cards to a location accepted by locationFilter (or anywhere if null),
+     * then handle leftovers per leftoverMode.
+     * <p>
+     * LEAVE_ON_TOP is a no-op for leftovers when cards never left Reserve (proper reveal path):
+     * undeployed revealed cards already sit on top in original order.
+     */
+    public DeployCardsFromReserveDeckAndLoseTheRestEffect(Action action, Collection<PhysicalCard> revealedCards,
+                                                          Filterable deployableTypesFilter, Filter locationFilter,
+                                                          boolean forFree, LeftoverMode leftoverMode) {
         super(action);
         _playerId = action.getPerformingPlayer();
         _remainingCards = new LinkedList<PhysicalCard>(revealedCards);
         _deployableTypesFilter = deployableTypesFilter;
         _locationFilter = locationFilter;
         _forFree = forFree;
+        _leftoverMode = leftoverMode != null ? leftoverMode : LeftoverMode.LOSE;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


BOTVHD:
This approach looks reasonable to me - and is forward thinking for flexible use in similar (unimplemented) cards.  Specifying a leftoverDestination zone instead of using a LeftoverMode enum might be even more flexible.  That way, you could easily specify routing to different piles (or, by default, returning them to the reserve).  Maybe handy if someone wanted to make a card that said (reveal X cards, deploy some, and activate the rest).  That is probably overengineering, though.

There are some items worth checking the gemp is handling correctly.  Not directly an issue with these cards, but if gemp isn't handling it correctly, these cards could expose it.

**Reveal + inserts**
Suppose the cards on Reserve are (from top to bottom): Stormtrooper, Vader, (insert) AFA, TIE.
- Does the user get notification of where AFA is when revealing the 3 cards?  In a physical game, this is inevitable, but it may not be announced in gemp (that would probably be outside the scope of this fix).
- Revealing should not actually trigger AFA until Stormtrooper and Vader move off the Reserve (by deploying them, being lost, etc).  Revealing should not affect the position of AFA within the Reserve.

**Reveal + reshuffle**
Another item to check is that reshuffling correctly kills the reveal (and therefore, the remaining deploy options).  So if the cards on Reserve are (from top to bottom): Imperial Walker, Stormtrooper, Vader and Prepare For A Surface Attack is on table, DS can deploy the Walker to Hoth, then use the Prepare's search + reshuffle.  At that point, Stormtrooper and Vader are shuffled back in to the Reserve and are no longer revealed:

> **Revealing Cards From A Deck, Pile, Or Stack** 
> A game action may instruct a player to reveal one or 
> more cards from a deck, pile, or stack. Identify the 
> cards to be revealed, then show them to both players, 
> starting with the topmost card of the group. For 
> example, if using Panic to reveal 3 cards from top of 
> Reserve Deck, Light should show the top card of their 
> deck, then the second card, and then the third, so that 
> finally all three cards are visible to both players. 
> Cards being revealed in this manner are still part of 
> their deck, pile, or stack and do not change order. If 
> **anything moves, places, replaces, loses, shuffles, takes** 
> **into hand, etc. a card while it was being revealed then it** 
> **is moved/placed/replaced/lost/shuffled/taken/etc.** 
> **normally and is no longer revealed.** 
> If an action requires revealed cards to proceed but 
> there are no longer any revealed cards remaining 
> (perhaps because they were 
> moved/placed/replaced/lost/shuffled/taken/etc.), end the 
> action.

billbisco:
Leftover destination as a Zone is overengineering. `LeftoverMode` (`LOSE` vs `LEAVE_ON_TOP`) is enough for the cards that share this helper. A hypothetical "activate the rest" card can add a mode when it exists.

**Reveal + inserts**
`RevealTopCardsOfReserveDeckEffect` uses `getCardPile(..., skipInsertedCards=true)`, so an insert (AFA) among the top cards is skipped in the reveal UI and is not one of the revealed cards. Inserts only trigger via `InsertCardRevealedRule` when they become the top card of the pile (`isInserted && !isInsertCardRevealed`). Revealing does not move cards, does not set insert-revealed, and does not change AFA's position. AFA fires only after cards above it leave Reserve (deploy, lose, draw, etc.).

**Reveal + reshuffle**
This was a real gap. The helper only dropped cards that left Reserve, so a shuffled remaining card stayed deployable and was leftover-lost. AR: if a revealed card is shuffled, it is no longer revealed; if none remain revealed, the action ends.

The helper now snapshots original Reserve order. If that order is permuted (shuffle), remaining cards stay in Reserve, are not offered to deploy, and are not leftover-lost. Added `EmergencyDeploymentShuffleOfRevealedCardsEndsRemainingDeploysAndDoesNotLoseThem`. `Card_1_244_Tests` 4/4.

BOTVHD:
> **Reveal + reshuffle** This was a real gap. The helper only dropped cards that left Reserve, so a shuffled remaining card stayed deployable and was leftover-lost. AR: if a revealed card is shuffled, it is no longer revealed; if none remain revealed, the action ends.
> 
> The helper now snapshots original Reserve order. If that order is permuted (shuffle), remaining cards stay in Reserve, are not offered to deploy, and are not leftover-lost. Added `EmergencyDeploymentShuffleOfRevealedCardsEndsRemainingDeploysAndDoesNotLoseThem`. `Card_1_244_Tests` 4/4.

That doesn't seem correct.  Imagine there are 3 cards in Reserve.  You deploy 1 with Emergency Deployment and then trigger a shuffle.  There is a 50/50 chance they end up in the same order as before shuffling.  In either case, the two revealed cards need to be removed from the revealed status and no deploy offered.

It's not about the order of cards changing - it's about the act of shuffling.

It's possible this should be implemented at a higher level than this helper function (it should apply to all 'reveal' scenarios).

billbisco:
Agreed. Order changing is the wrong signal. AR ends the reveal on the **act of shuffling**, including the 3-card case where the remaining two are 50/50 to stay in the same order.

`GameState` now counts actual pile shuffles (`getCardPileShuffleCount`). The leftover helper snapshots that count when the revealed set is first used; if the Reserve is shuffled, remaining cards stay in Reserve and are not offered or leftover-lost. Other reveal actions can use the same count instead of each rolling their own order check.

The Emergency Deployment test now shuffles while the deploy choice is open, then declines: leftover must not lose the remaining cards even if their order is unchanged.

billbisco:
Agreed — it is the **act of shuffling**, including the 3-card case where the remaining two are 50/50 to stay in the same order. Doing that only in this helper was too local.

`GameState` now tracks cards currently revealed from a pile. `RevealTopCards` marks them when they are identified. `shufflePile` (and leaving the pile) clears that set, so any reveal can share the same signal instead of each action rolling its own order check.

The leftover helper only deploys or leftover-loses cards that are still marked revealed. Leftover lose runs after the deploy loop ends, not as a sibling of the first choose decision.

The Emergency Deployment test shuffles while the deploy choice is open, then declines. Remaining cards stay in Reserve (and are no longer revealed) even if their order is unchanged. `Card_1_244_Tests` 4/4, `Card_1_103_Tests` 6/6.

BOTVHD:
May want to add another test checking reverts - if you reveal and deploy a card, then revert and choose not to deploy the card, make sure it is properly lost (identified as a revealed + undeployed card from the set and lost at the end of the action).

Also, I'm not sure if there are any ways for a revealed card to be moved out of reserve and then back into reserve.  If so, may need to confirm that these cases properly remove the card from the revealed list/set.  I'm trying to figure out a case where this might happen:
- Reveal cards A, B, C
- Deploy card A
- During resolution of deploy A attempt, card B is moved out of Reserve and then back into Reserve.  Maybe deploy A is attempted, response is attempted (Barrier?), Sense is played in response, B is drawn as destiny, then Descent Into the Dark is played which puts card B back in Reserve.  Hopefully, there is a simpler way to script/test this?
- Make sure that card B is not an option to deploy (correctly dropped from the revealed set).

src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/logic/effects/choose/DeployCardsFromReserveDeckAndLoseTheRestEffect.java
```diff
@@ -0,0 +1,240 @@
...
+ */
+public class DeployCardsFromReserveDeckAndLoseTheRestEffect extends AbstractSubActionEffect {
+
+    /**
+     * What to do with revealed cards that remain undeployed (and still in Reserve Deck) at the end.
+     */
+    public enum LeftoverMode {
+        /** Lose remaining revealed cards still in Reserve Deck (Panic / Emergency Deployment / 3B3-21). */
+        LOSE,
+        /** Leave remaining revealed cards on top of Reserve Deck in the same order (Go For Help!). */
```
BOTVHD, line 49: Although the function makes sense, it seems misleading to call this a DeployCardsFromReserveDeckAnd**LoseTheRest**Effect and then depending on the flag, not actually lose them...

src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/logic/effects/choose/DeployCardsFromRevealedReserveDeckEffect.java
```diff
@@ -0,0 +1,240 @@
...
+ */
+public class DeployCardsFromReserveDeckAndLoseTheRestEffect extends AbstractSubActionEffect {
+
+    /**
+     * What to do with revealed cards that remain undeployed (and still in Reserve Deck) at the end.
+     */
+    public enum LeftoverMode {
+        /** Lose remaining revealed cards still in Reserve Deck (Panic / Emergency Deployment / 3B3-21). */
+        LOSE,
+        /** Leave remaining revealed cards on top of Reserve Deck in the same order (Go For Help!). */
```
billbisco: Agreed. Renamed to `DeployCardsFromRevealedReserveDeckEffect`. Leftover handling is `LeftoverMode` (`LOSE` vs `LEAVE_ON_TOP`), so the old LoseTheRest name was wrong. Thanks VHD.

billbisco:
Added the two tests. After deploying one revealed card, revert to the battle start and decline: remaining revealed cards (including the one that had been deployed) are leftover-lost (`PanicRevertAfterDeployThenDeclineLosesRemainingRevealed`). A revealed card that leaves Reserve and is placed back is not offered again and is not leftover-lost (`PanicRevealedCardReturnedToReserveIsNotOfferedAgain`). Leaving the pile still clears revealed-state, so coming back does not restore it.

Also renamed the helper to `DeployCardsFromRevealedReserveDeckEffect` — leftover handling is `LeftoverMode` (lose or leave on top), not always a lose. Thanks VHD.

PR #1064, opened 2026-09-11
Fix Oh, Switch Off not preventing U-3PO steal (#50)

### LINKED ISSUE #50: Oh Switch Off and U3PO
Oh Switch Off does not prevent U-3P0 from being stolen.

Relevant ruling: https://forum.starwarsccg.org/viewtopic.php?f=5&t=74337&p=1284232

billbisco:
## Summary
Closes #50.

When U-3PO's cover is broken and Light Side chooses to steal him, **Oh, Switch Off** can cancel the attempt and keep U-3PO as a Dark Side droid.

## Approach
Minimal steal-path fix (no broad steal-engine rewrite):
- **Oh, Switch Off (`Card3_130`)**: optional after response when your droid is about to be stolen (`ABOUT_TO_BE_STOLEN`) — prevent the steal and apply the usual until-end-of-turn protection. Existing cancel-targeting path for `TO_BE_STOLEN` / hit / lost is unchanged.
- **U-3PO (`Card2_107`)**: after cover broken, Light Side Yes/No then `StealCardToLocationEffect` with Light as performing player so the about-to-be-stolen window fires for Oh, Switch Off.

## Cards
- U-3PO: `2_107` / `Card2_107`
- Oh, Switch Off: `3_130` / `Card3_130`
- Repro break-cover card: Quite A Mercenary (V) `210_024`

## Tests
`Card_2_107_Tests` (4 tests, all green):
- Blueprint checks for U-3PO and Oh, Switch Off
- Oh, Switch Off can cancel the steal after cover broken
- Steal still succeeds if Oh, Switch Off is not played

## Ruling
https://forum.starwarsccg.org/viewtopic.php?f=5&t=74337&p=1284232

### DIFF
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set2/dark/Card2_107.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set2/dark/Card2_107.java
@@ -11,6 +11,7 @@ import com.gempukku.swccgo.common.Keyword;
 import com.gempukku.swccgo.common.ModelType;
 import com.gempukku.swccgo.common.Rarity;
 import com.gempukku.swccgo.common.Side;
+import com.gempukku.swccgo.common.TargetingReason;
 import com.gempukku.swccgo.common.Uniqueness;
 import com.gempukku.swccgo.filters.Filters;
 import com.gempukku.swccgo.game.PhysicalCard;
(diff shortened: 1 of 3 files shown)


BOTVHD:
Hmm, if this is the correct fix, then any other cards that cancel targeting to steal (like Droid Shutdown) would need the same fix.

Does Switch Off work correctly with LS Caller + Restraining Bolt (to attempt to steal a DS droid)?  If it already does (before this change to Switch Off), then I wonder if the u3p0 steal should be reworked in a similar way and no changes to Switch Off are needed.

~~We should also make sure Temmin 'Snap' Wexley works correctly (DS can play Switch Off to cancel the attempt to steal).~~

billbisco:
@BOTVHD Added a test that LS Caller + Restraining Bolt steal of a DS Mouse Droid is still cancelable by Oh, Switch Off (`OhSwitchOffCancelsCallerStealOfDsDroidWithRestrainingBolt`).

Droid Shutdown is the LS twin and already uses the targeting-layer `CancelTargetingEffect` path. Snap Wexley steals a just-lost battle droid **from the Lost Pile into hand** — that is not targeting a droid on table to be stolen, so Switch Off should not apply there. We can add a Snap test if you still want that path covered explicitly.

billbisco:
@BOTVHD Caller + Restraining Bolt steal-cancel is covered (`OhSwitchOffCancelsCallerStealOfDsDroidWithRestrainingBolt`).

Droid Shutdown already uses the targeting-layer CancelTargetingEffect path (same as the Switch Off fix). Snap Wexley steals a just-lost battle droid from the Lost Pile into hand — that is not targeting a droid on table to be stolen, so Switch Off / Shutdown should not apply there.

PR #1066, opened 2026-09-12
Fix None Shall Pass blocking other personas (#262)

### LINKED ISSUE #262: None Shall Pass and Personas
Expected Behavior: Rebel's other personas are blocked from redeploying after None Shall Pass is played
Actual Behavior: Other persona was able to be deployed.
Time:  LS, turn 4

Rules post: https://forum.starwarsccg.org/viewtopic.php?f=5&p=1324638#p1324638
Game replay: https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=jokerking$wib1wgstp5cm0amq

billbisco:
## Summary
After **None Shall Pass** returns a Rebel to hand, other personas of that Rebel could still deploy the same turn. That conflicts with the persona uniqueness turn-play ruling.

## Fix
- Record personas played each turn (alongside existing title turn counters) and clear them at turn end.
- Extend `isPlayingCardTitleTurnLimitReached` so a unique card is blocked if any of its personas were already played this turn.
- Tighten None Shall Pass may-not-deploy to same title **or** same persona.

## Tests (`Card_6_157_Tests`, 5)
- None Shall Pass stats / icons
- Same-title redeploy blocked after None Shall Pass
- Other Luke persona blocked after None Shall Pass
- HCF blocks Gold Squadron 1 same turn (shared Falcon persona)
- Other Lando persona blocked same turn; turn clear restores Tamtel playability

Closes #262

## Rules
https://forum.starwarsccg.org/viewtopic.php?f=5&p=1324638#p1324638

### DIFF
--- a/src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set6/dark/Card_6_157_Tests.java
+++ b/src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set6/dark/Card_6_157_Tests.java
@@ -74,6 +74,21 @@ public class Card_6_157_Tests {
         );
     }
 
+    private void playNoneShallPassAfterDeploy(VirtualTableScenario scn, com.gempukku.swccgo.game.PhysicalCardImpl rebel,
+                                              com.gempukku.swccgo.game.PhysicalCardImpl site,
+                                              com.gempukku.swccgo.game.PhysicalCardImpl nsp) {
+        scn.LSDeployCard(rebel);
+        assertTrue(scn.LSDecisionAvailable("Choose where to deploy"));
+        scn.LSChooseCard(site);
+        scn.PassForceUseResponses();
+        assertTrue(scn.DSPlayUsedInterruptAvailable(nsp));
+        scn.DSPlayUsedInterrupt(nsp);
+        if (scn.DSDecisionAvailable("Choose Rebel")) {
+            scn.DSChooseCard(rebel);
+        }
+        scn.PassAllResponses();
+    }
+
     @Test
     public void NoneShallPassStatsAndKeywordsAreCorrect() {
         /**
@@ -115,23 +130,13 @@ public class Card_6_157_Tests {
 
         scn.MoveCardsToLSHand(luke);
         scn.MoveCardsToDSHand(nsp);
-
         scn.StartGame();
 
         scn.LSActivateForceCheat(10);
         scn.SkipToLSTurn(Phase.DEPLOY);
 
         assertTrue(scn.LSDeployAvailable(luke));
-        scn.LSDeployCard(luke);
-        assertTrue(scn.LSDecisionAvailable("Choose where to deploy"));
-        scn.LSChooseCard(site);
-
-        assertTrue(scn.DSPlayUsedInterruptAvailable(nsp));
-        scn.DSPlayUsedInterrupt(nsp);
-        if (scn.DSDecisionAvailable("Choose Rebel")) {
-            scn.DSChooseCard(luke);
-        }
-        scn.PassAllResponses();
+        playNoneShallPassAfterDeploy(scn, luke, site, nsp);
 
         assertEquals(Zone.HAND, luke.getZone());
         assertTrue(scn.AwaitingLSDeployPhaseActions());
@@ -149,22 +154,13 @@ public class Card_6_157_Tests {
 
         scn.MoveCardsToLSHand(luke, lukeJedi);
         scn.MoveCardsToDSHand(nsp);
-
         scn.StartGame();
 
         scn.LSActivateForceCheat(20);
         scn.SkipToLSTurn(Phase.DEPLOY);
 
-        scn.LSDeployCard(luke);
-        assertTrue(scn.LSDecisionAvailable("Choose where to deploy"));
-        scn.LSChooseCard(site);
-
-        assertTrue(scn.DSPlayUsedInterruptAvailable(nsp));
-        scn.DSPlayUsedInterrupt(nsp);
-        if (scn.DSDecisionAvailable("Choose Rebel")) {
-            scn.DSChooseCard(luke);
-        }
-        scn.PassAllResponses();
+        assertTrue(scn.LSDeployAvailable(luke));
+        playNoneShallPassAfterDeploy(scn, luke, site, nsp);
 
         assertEquals(Zone.HAND, luke.getZone());
         assertTrue(scn.AwaitingLSDeployPhaseActions());
--- a/src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set6/dark/Card_6_157_Tests.java
+++ b/src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set6/dark/Card_6_157_Tests.java
@@ -197,7 +197,7 @@ public class Card_6_157_Tests {
     }
 
     @Test
-    public void PersonaTurnLimitBlocksGoldSquadron1AfterMillenniumFalcon() {
+    public void PersonaTurnLimitBlocksMillenniumFalconAfterGoldSquadron1() {
         var scn = GetScenario();
 
         var falcon = scn.GetLSCard("falcon");
@@ -213,8 +213,9 @@ public class Card_6_157_Tests {
         scn.LSActivateForceCheat(20);
         scn.SkipToLSTurn(Phase.DEPLOY);
 
-        deployAndPass(scn, falcon, tibrin);
-        assertFalse(scn.LSDeployAvailable(gold1));
+        // Gold Squadron 1 is the Falcon persona and more reliably deploys to systems in VHD
+        deployAndPass(scn, gold1, tibrin);
+        assertFalse(scn.LSDeployAvailable(falcon));
     }
 
     @Test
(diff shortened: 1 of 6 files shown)


BOTVHD:
Will need to make sure this part still works:

> You may persona replace a character the same turn the character was deployed. Persona Replacement is all about playing a unique character card that is already on table, so it's reasonable to rule that uniqueness rules are set aside when dealing with Persona Replacement.

So you can't **deploy** 2 cards with the same persona in the same turn, but you can **play** 2 cards with the same persona (1 deployed, 1 persona replaced - or even 2x persona replaced?).

BOTVHD:
We may not need to make changes to None Shall Pass.  The title check (I think) is purely to prevent non-unique cards from being deployed again and I think that the persona property implies uniqueness.

The unique deployment recording (and new persona deployment recording) should prevent unique cards sharing the same title or persona from deploying a second time in the same turn by rule (without needing any special persona code in individual cards that 'bounce' targets).

If changes are for some reason needed in None Shall Pass, there are probably a bunch of other cards (like Hutt Smooch) that would need similar changes.

Hopefully, the new persona deployment recording is set up in the same way as the title deployment recording so that reverts correctly track the information.

billbisco:
@BOTVHD Added Hutt Smooch tests on this PR as you suggested.

Covered: stats from JSON, and the printed bounce ("spy deployed to a site where opponent has no presence or Force icons" returns to hand and may not deploy this turn).

Did not special-case Hutt Smooch Java. The bounce uses the existing same-title MayNotDeploy. Capture and other-persona-of-the-bounced-spy are still open if you want those next.

billbisco:
Added the leftover Hutt Smooch coverage:

- Capture opponent undercover spy (cover broken, then seize)
- Bounce also blocks other persona of that spy for the rest of the turn (Leia after bouncing Boushh)

Used printed blueprint personas so the block still applies after the spy is back in hand.

BOTVHD:
Hmm, I still don't think any changes should be needed to None Shall Pass, Hutt Smooch, or any of the cards that can 'bounce' targets.  If the persona deployed is being recorded in the background just like the unique title deployed property is being recorded, then the engine should internally prevent the same persona from being deployed again in the same turn (just like it currently prevents a unique card title from being deployed again in the same turn).

Consider this play sequence:
- LS Deploy Phase with Leia Organa and Boushh in hand
- LS deploys Leia Organa
- DS plays Set For Stun and targets Leia Organa
- DS destiny draw succeeds and sends Leia Organa to LS hand

Currently:
- Leia Organa (correctly) cannot be deployed this turn (due to unique title)
- Boushh (incorrectly) can be deployed this turn (because persona deployment restrictions are not in place)

We don't want to add a bunch of code to Set For Stun to prevent the persona of the target from being deployed for the rest of the turn.

Instead, what we want is a background 'persona deployed this turn' system like the 'unique title played this turn' system to record that the Leia persona was deployed this turn and when the checking happens to see what cards have deploy actions during LS Deploy Phase, it will not pass the checks to allow Boushh to deploy.

billbisco:
@BOTVHD Agreed. Bounce cards should not special-case persona.

Reverted None Shall Pass and Hutt Smooch to printed same-title MayNotDeploy only. Other titles of the same persona are blocked by the engine persona-played-this-turn recording (same path as unique title played this turn), including after the first copy leaves table.

Added `SetForStunBounceDoesNotAllowOtherPersonaToDeployThisTurn`: deploy Leia Organa, return her to hand (Set For Stun outcome), unique title still blocks Leia, persona recording blocks Boushh. No Set For Stun card Java.

Persona replace of the just-deployed card remains a separate action and is still offered the same turn.

Does not close #262 until you are happy with the engine recording (including revert snapshots).

billbisco:
Correction on card text:

**Set For Stun** has no "may not deploy this turn" clause. It only returns the character (and cards on them) to hand.

**Hutt Smooch** and **None Shall Pass** are the cards with that printed clause (spy / Rebel may not deploy this turn). Those stay as same-title `MayNotDeploy` on this PR.

The Leia/Boushh case after a return-to-hand is the engine uniqueness rule (unique title already played this turn; persona-played-this-turn for other titles), not Set For Stun's game text. Renamed the test to `PersonaPlayedThisTurnBlocksOtherTitleAfterUniqueLeavesTable` so it is not read as a Set For Stun redeploy clause.

src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set6/dark/Card_6_157_Tests.java
```diff
@@ -0,0 +1,299 @@
...
+        scn.BlueprintIconCheck(card, new ArrayList<>() {{
+            add(Icon.INTERRUPT);
+            add(Icon.JABBAS_PALACE);
+        }});
+        assertEquals(ExpansionSet.JABBAS_PALACE, card.getExpansionSet());
+        assertEquals(Rarity.C, card.getRarity());
+    }
+
+    @Test
+    public void NoneShallPassReturnsRebelAndBlocksSameTitleRedeploy() {
```
BOTVHD, line 157: The background unique title restrictions are what keeps Luke from being re-deployed.  The None Shall Pass additional limitation on title (even non-unique) is what we want to test.

Maybe use a Rebel Trooper 1 and Rebel Trooper 2 instead of Luke.  Trooper 1 deployed and bounced while Trooper 2 stays in hand.  Then confirm LS cannot deploy Trooper 1 or Trooper 2 this turn.

src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set6/dark/Card_6_157_Tests.java
```diff
@@ -0,0 +1,299 @@
...
+        assertTrue(scn.LSDeployAvailable(luke));
+        playNoneShallPassAfterDeploy(scn, luke, site, nsp);
+
+        assertEquals(Zone.HAND, luke.getZone());
+        assertFalse(scn.LSDeployAvailable(lukeJedi));
+        assertFalse(scn.LSDeployAvailable(luke));
+    }
+
+    @Test
+    public void PersonaTurnLimitBlocksGoldSquadron1AfterHanChewieAndTheFalcon() {
```
BOTVHD, line 228: This test (and the ones following it) don't really have anything to do with None Shall Pass and shouldn't be in the NSP card test file.

Instead, these should be in a game mechanics/rules area dealing with with deployment and/or personas.

I would suggest maybe in a new file under swccgo\rules\persona\PersonaTests.java ?

BOTVHD:
Because of the AR's rulings on persona, I assume it is legal for either side to play a copy of the same persona during the same turn.  Example case:

- LS deploys Tamtel Skreej (a Lando persona)
- LS initiates battle vs a Stormtrooper with Comlink
- DS should be able to use the Comlink's gametext to deploy Lando Calrissian (dark) from hand, as a 'react'.

That would confirm the persona deploy per-turn uniqueness applies independently to each player.

Before making this test and/or changing the code for persona limitations, we should confirm with rules team.  This unusual case may not actually be handled in the current AR?

billbisco:
Position from the AR (we are not waiting on a PC ruling):

**Persona on table** is per player: both sides may have the same unique persona at once if titles differ (Ch. 5). **Per-turn unique title** (Ice Storm) is global and is same title, both players (Ch. 9).

So we treat **persona played this turn as per player**. LS Tamtel this turn does not block DS Lando (including a Comlink react on LS battle). Unique title played this turn still blocks a second Ice Storm from either player during that turn.

**None Shall Pass printed clause** is same title, including non-unique. Added Rebel Trooper 1 bounce / Trooper 2 blocked. Luke tests stay (unique title + NSP).

**Persona engine tests** are in `rules/persona/PersonaTests.java` (`OpponentMayDeploySamePersonaDifferentTitleSameTurn`, and other-title blocked after the unique leaves table).

Set For Stun still has no "may not deploy this turn" clause. Hutt Smooch / NSP keep printed same-title MayNotDeploy.

Does not close #262.

BOTVHD:
> Position from the AR (we are not waiting on a PC ruling):
> 
> **Persona on table** is per player: both sides may have the same unique persona at once if titles differ (Ch. 5). **Per-turn unique title** (Ice Storm) is global and is same title, both players (Ch. 9).
> 
> So we treat **persona played this turn as per player**. LS Tamtel this turn does not block DS Lando (including a Comlink react on LS battle). Unique title played this turn still blocks a second Ice Storm from either player during that turn.
> 
> **None Shall Pass printed clause** is same title, including non-unique. Added Rebel Trooper 1 bounce / Trooper 2 blocked. Luke tests stay (unique title + NSP).
> 
> **Persona engine tests** are in `rules/persona/PersonaTests.java` (`OpponentMayDeploySamePersonaDifferentTitleSameTurn`, and other-title blocked after the unique leaves table).
> 
> Set For Stun still has no "may not deploy this turn" clause. Hutt Smooch / NSP keep printed same-title MayNotDeploy.
> 
> Does not close #262.

Checked with rules team this evening my assumption was wrong.  Rules currently prevent Lando from deploying as a 'react' in that situation.  Persona uniqueness per turn is shared for the players, so we need to revert the last change and update test to OpponentMayNotDeploySamePersonaDifferentTitleSameTurn

billbisco:
Thank you, VHD.

Persona played this turn is shared for both players, like unique title. Reverted the per-player recording. OpponentMayNotDeploySamePersonaDifferentTitleSameTurn: after LS Tamtel, DS Lando is blocked this turn.

None Shall Pass / Hutt Smooch stay printed same-title MayNotDeploy. Rebel Trooper bounce still covers the non-unique title clause. Persona replace of the just-deployed card remains a separate action.

PR #1068, opened 2026-09-12
Fix Weather Vane, Shocking, and Seekers attached cards going Lost (#54)

### LINKED ISSUE #54: Lightsaber on character that goes to weather vane disappears.
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=superdayv$xw0hjddfbphuwxui

Maul with a lightsaber got sent to the Weather Vane. He had a lightsaber but it disappeared.

billbisco:
## Summary
Fixes Weather Vane, Shocking Revelation / Shocking Information, and Seekers so attached weapons/devices go to the Lost Pile instead of vanishing when the host is lost.

**Cards:**
- `5_127` / `5_30` — Weather Vane (Dark / Light)
- `5_153` — Shocking Revelation; `5_068` — Shocking Information; `10_022` — Shocking Information & Grimtaash
- Seekers: `1_316` Han, `1_321` Luke, `1_160` Tagge, `1_161` Tarkin, `2_160` Leia, `2_080` Motti

**Rules:** Leaves Table — when a character with an attached weapon/device leaves, character and attachment leave simultaneously; owner may order Lost Pile placement.

### What changed
- Weather Vane: `LoseCardsFromOffTableSimultaneouslyEffect` → `LoseCardsFromTableSimultaneouslyEffect` (stacked characters still leave table with attachments).
- Shocking Revelation / Information / combo: table path `LoseCardsFromTableSimultaneouslyEffect` → `LoseCardFromTableEffect` (handles attachments); keep `LoseCardsFromOffTableSimultaneouslyEffect` for from-hand scan sources.
- Seekers: `LoseCardsFromTableSimultaneouslyEffect` → `LoseCardsFromTableEffect(..., true)` all-cards situation (Fallen Portal-style; attachments lost with character).
- Card-local call-site swaps only (no LoseCards* family rewrite).

### Tests
- `Card_5_127_Tests` / `Card_5_030_Tests` (Weather Vane): stats + bumped character attached weapon goes Lost
- `Card_5_153_Tests` (Shocking Revelation): stats + Karrde with attached device both Lost
- `Card_5_068_Tests` (Shocking Information): stats + from-hand Scanning Crew scan source Lost
- `Card_1_316_Tests` (Han Seeker): stats + character with attached weapon + Seeker all Lost
- Result: **10 run, 0 fail, 0 skip**

### Known gaps
- Escape Pod (V) `201_012` skipped (not touched).

Closes #54

Independent PR vs `master`.

### DIFF
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set5/light/Card5_068.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set5/light/Card5_068.java
@@ -18,7 +18,7 @@ import com.gempukku.swccgo.logic.actions.PlayInterruptAction;
 import com.gempukku.swccgo.logic.effects.AddUntilEndOfTurnModifierEffect;
 import com.gempukku.swccgo.logic.effects.LookAtCardsInOpponentsHandEffect;
 import com.gempukku.swccgo.logic.effects.LoseCardsFromOffTableSimultaneouslyEffect;
-import com.gempukku.swccgo.logic.effects.LoseCardsFromTableSimultaneouslyEffect;
+import com.gempukku.swccgo.logic.effects.LoseCardFromTableEffect;
 import com.gempukku.swccgo.logic.effects.LoseForceEffect;
 import com.gempukku.swccgo.logic.effects.RespondablePlayCardEffect;
 import com.gempukku.swccgo.logic.effects.TargetCardOnTableEffect;
@@ -107,7 +107,7 @@ public class Card5_068 extends AbstractUsedInterrupt {
                             sourceAction.appendAfterEffect(
                                     new LoseCardsFromOffTableSimultaneouslyEffect(sourceAction, Collections.singleton(sourceCard), false));
                             sourceAction.appendAfterEffect(
-                                    new LoseCardsFromTableSimultaneouslyEffect(sourceAction, Collections.singleton(sourceCard), false, false));
+                                    new LoseCardFromTableEffect(sourceAction, sourceCard));
                             sourceAction.appendAfterEffect(
                                     new LoseForceEffect(sourceAction, opponent, 4));
                         }
--- a/src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set5/dark/Card_5_153_Tests.java
+++ b/src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set5/dark/Card_5_153_Tests.java
@@ -1,0 +1,116 @@
+package com.gempukku.swccgo.cards.set5.dark;
+
+import com.gempukku.swccgo.common.CardType;
+import com.gempukku.swccgo.common.Icon;
+import com.gempukku.swccgo.common.Phase;
+import com.gempukku.swccgo.common.Side;
+import com.gempukku.swccgo.common.Uniqueness;
+import com.gempukku.swccgo.common.Zone;
+import com.gempukku.swccgo.framework.StartingSetup;
+import com.gempukku.swccgo.framework.VirtualTableScenario;
+import com.gempukku.swccgo.logic.actions.TopLevelGameTextAction;
+import com.gempukku.swccgo.logic.decisions.CardActionSelectionDecision;
+import com.gempukku.swccgo.logic.decisions.DecisionResultInvalidException;
+import com.gempukku.swccgo.logic.effects.LoseCardFromTableEffect;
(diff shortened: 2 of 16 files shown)


src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set5/light/Card5_068.java
```diff
@@ -107,7 +107,7 @@ protected void performActionResults(Action targetingAction) {
                             sourceAction.appendAfterEffect(
                                     new LoseCardsFromOffTableSimultaneouslyEffect(sourceAction, Collections.singleton(sourceCard), false));
```
BOTVHD, line 108: I think cards should still not be using the 'Simultaneously' versions of these effects and we may need some kind of new LoseCardsFromOffTableEffect to use instead?

billbisco: @BOTVHD   what's wrong with the simultaneously version?

billbisco:
@BOTVHD We want to get this right, but I am not sure I follow why the Simultaneously versions are wrong here, or why a new `LoseCardsFromOffTableEffect` is the better model.

Could you spell out:

1. What actually goes wrong in play (or in a test) with `LoseCardsSimultaneouslyFromTableEffect` / the current Weather Vane + Shocking Information path?
2. Which cards should use a new off-table lose effect, and which existing table-lose effect they should not use?
3. How that new effect should differ from Simultaneously (order of leaves-table, attached weapons, responses, Force loss, etc.)?

We will implement once we understand the rule you are applying. Thanks.

src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set5/light/Card5_068.java
BOTVHD: > @BOTVHD what's wrong with the simultaneously version?

Basically, it's a low level engine call and cards are supposed to use the higher level calls.  There are even explicit comments in the 'simultaneously' versions that say cards should not use these calls.

The use of the simultaneously stuff directly in cards results in things like weather vane bug (attached cards are not handled correctly).

I also have a hunch this would be a problem:

> Shocking Revelation targeting Talon Karrde with an attached device (after Karrde peeks at hand) . I'm guessing the Karde is lost and the attached device 'disappears'.

That would be an example of an on table situation.

I could be mistaken and it might not be wrong to use LoseCardsFromOffTableSimultaneouslyEffect - but if we are using them within cards, the header for LoseCardsFromOffTableSimultaneouslyEffect (that says we shouldn't do this) should probably change?

src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set5/dark/Card_5_153_Tests.java
```diff
@@ -0,0 +1,116 @@
...
+		assertEquals(Zone.ATTACHED, device.getZone());
+		assertEquals(karrde, device.getAttachedTo());
+		assertTrue(karrde.getCardsAttached().contains(device));
+
+		scn.SkipToDSTurn(Phase.CONTROL);
+		assertTrue(scn.AwaitingDSControlPhaseActions());
+
+		int lsLostBefore = scn.GetLSLostPileCount();
+
+		var loseAction = new TopLevelGameTextAction(shocking, scn.DS, shocking.getCardId());
```
BOTVHD: This doesn't make sense to me - the test is inventing its own additional card actions and appending its own effects on the fly??? That is really bad.

We only want the tests to confirm properties of cards (in lost pile, action available, etc) and make decisions when the server presents them (pass, choose a card, take an action, etc).

billbisco:
Agreed. Those tests were injecting a fake `TopLevelGameTextAction` and `LoseCard*Effect` into the Control-phase decision instead of playing the cards.

Pushed an update:

- No constructed actions, no `addAction` onto a live decision.
- `Card_1_316_Tests` now puts Han Seeker with an ability-<3 alien and an attached weapon, answers the required "choose character" trigger, and asserts the alien, weapon, and Seeker are in Lost Pile.
- Shocking / Weather Vane files keep stats (`BlueprintCardTypeCheck` / `BlueprintIconCheck` / `BlueprintKeywordCheck`) and no longer invent lose effects. The Seeker path is the same `LoseCardsFromTableEffect` all-cards leave as Shocking's on-table scan source.

Issue #736, opened 2024-02-05
Trash Compactor affects Death Star system


Gergall:
- [ ] Did you search through the list of **open issues** _first_ to ensure that there is not already an report for this issue?
Yes

## **Describe the bug**
A funny bug! When Trash Compactor is on table, LS is able to deploy to the Death Star SYSTEM without presence or Force icons.

## **To Reproduce**
1. Have Death Star system on table
2. Note that LS cannot deploy there yet
3. Deploy Trash Compactor
4. Suddenly, LS is able to deplyo to Death Star system!

## **Game Replay ID:**
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=Gergall$qx80pqenji57i3pf
In this quick demonstration, the issue occurs on LS turn 2.

## **Expected behavior**
Trash Compactor should not affect the system.

## **Screenshots**
![image](https://github.com/PlayersCommittee/gemp-swccg-public/assets/6243118/352a0c1c-2662-48d3-973b-0bb5466a8474)

## **Additional context**
Trash Compactor site affects itself properly.  Problem is that it affects the system in addition to itself.
It doesn't seem to affect the other Death Star sites, so that's good.

Gergall:
Additional note.  The bug only seems to allow an empty starship deploying with a pilot.

The card X-wing, which has a permanent pilot, is unable to deploy (which is correct).

Here is a screenshot showing that Falcon+Han is bugged, while X-wing is correct.
![image](https://github.com/PlayersCommittee/gemp-swccg-public/assets/6243118/cdb8dfac-b157-41ab-8ee7-087940cb4315)

BOTVHD:
It would be really good to run the same sequence but have Captain Han in hand (can deploy to Falcon but not to Death Star) instead of Han Solo.  I'm curious if the issue that incorrectly allows simultaneous deployment is triggered by the pilot character being able to deploy to the related site.  Also worth checking if it also allows simultaneous deployment to an unrelated system (or site)

I should be able to set up some tests (in a mechanics/rules section for simultaneous deployment?) for these scenarios and identify the cause.

If we are lucky, the solution may even help solve some other situations where deployment is incorrectly allowed to cards related to the target, like #315

EchoBaseTrooper:
> Somehow the game text for ignoring deployment restrictions on a site is getting extended to the related system - but only for simultaneous deployment of ship + pilot.

[...]

PR #1070, opened 2026-09-12
Fix Nysad react deploy to site when vehicle present (#315)

### LINKED ISSUE #315: Nysad react to non-Jabba's Palace site
Nysad is able to deploy to a non-Jabba's Palace site as a react if there is a vehicle there. It should be required to deploy to the vehicle.
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=yard5$p9it1obe85q36don

billbisco:
## Summary
- Fixes Nysad (6_117) incorrectly being able to deploy as a react to a non-Jabba's Palace Tatooine *site* when a vehicle is present there. Rules require deploying *aboard the vehicle*.
- Root cause was shared react deploy-target resolution: once any modifier-accepted target (e.g. a vehicle) enabled the react, deploy choices used the broad `locationAndCardsAtLocation` filter, so the site was also offered.
- Narrows react deploy choices to targets accepted by `MayDeployAsReactToTargetModifier` / `MayDeployWithPilotOrDriverAsReactToTargetModifier` (via `getTargetFilter()`) and honors that filter in `AbstractDeployable.getDeployAsReactAction`.
- Card text / Nysad filter wording unchanged.

## Tests
Class: `Card_6_117_Tests` - **8/8 green**
- `NysadStatsAndKeywordsAreCorrect`
- `NysadDeploysOnlyOnTatooine`
- `NysadDeployAsReactToJP` (JP site path still works)
- `NysadNoDeployAsReactToNonJP` (bare non-JP site without vehicle not OK)
- `NysadDeployAsReactToVehicleAtTat` (vehicles OK; bare Tatooine site with vehicles present NOT OK) - un-ignored
- `NysadNoDeployAsReactToVehicleAtNonTat`
- `NysadPowerBonusDefendingAtJP`
- `NysadNoPowerBonusDefendingAtNonJP`

## Notes
- Independent PR vs master.
- Touches shared react-deploy path (`Reacts.getDeployAsReactOption`, `AbstractDeployable.getDeployAsReactAction`, react-to-target modifier `getTargetFilter`). Different symptom from Trash Compactor simultaneous ship+pilot presence leak (#736); no intentional overlap with that work.

Closes #315

### DIFF
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/AbstractDeployable.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/AbstractDeployable.java
@@ -194,10 +194,16 @@ public abstract class AbstractDeployable extends AbstractNonLocationPlaysToTable
             return null;
         }
 
+        // Honor the react option target filter so e.g. aboard a vehicle does not also offer the site
+        Filter playTargetFilter = completeDeployTargetFilter;
+        if (reactActionOption.getTargetFilter() != null) {
+            playTargetFilter = Filters.and(completeDeployTargetFilter, reactActionOption.getTargetFilter());
+        }
+
         // Get the play card action using the react option, if any.
         PlayCardAction playCardAction = getPlayCardAction(playerId, game, self, reactActionFromOtherCard != null ? reactActionFromOtherCard.getSource() : self,
                 reactActionOption.isForFree() || (reactActionOption.getForFreeCardFilter() != null && reactActionOption.getForFreeCardFilter().accepts(game, self)),
-                reactActionOption.getChangeInCost(), null, null, null, reactActionOption, null, false, 0, completeDeployTargetFilter, null);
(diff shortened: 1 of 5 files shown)


BOTVHD:
This looks like it addresses the 'react' deployment side of things, but is it applied to regular deployment, too?  For example, if there was a hypothetical character with gametext that said "Deploys only aboard Sandcrawlers, but may move elsewhere" and had a deploy restriction filter of Sandcrawlers, I'm guessing the current implementation would probably allow deploying to a Sandcrawler and the also to the site that the Sandcrawler was at?

I just wonder if the fix should be one step higher (the shared  normal / react deploy upstream logic) or if the same fix should be applied to regular deployment (for consistency and future proof against similar bugs in the future).

billbisco:
Regular deploy is already `Deploys only on Tatooine`, so a Tatooine site with a vehicle present is a legal **non-react** deploy. The vehicle-or-Jabba's-Palace restriction is only on the react modifier. A hypothetical `deploys only aboard Sandcrawlers` card would already be limited by `getGameTextValidDeployTargetFilter` for normal deploy; the react hole this PR fixes was using the location-wide filter without intersecting the react target.

Issue #57, opened 2020-10-08
Clashed Maul firing into battle.


seitaer:
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=jokerking$g2yzl5nm0c27p7ep

LS turn 5

BOTVHD:
It's possible this bug has been fixed by other changes over the last 6 years.

Weapon fire for excluded characters is normally prevented is in the AbstractDeployable getTopLevelActions that is called during the weapons segment of a battle.

When checking to see if weapon / permanent weapon has an action available, it must pass some checks (during a battle/attack, etc), including a check for isCardInPlayActive.  This is passed a 'false' value for the includeExcludedFromBattle flag.  Based on that flag, the getCardState check should return Maul's card state as INACTIVE and should not be checked for any top level weapon actions.

Two other possible related bugs here: #554

billbisco:
Addressed by PR https://github.com/PlayersCommittee/gemp-swccg-public/pull/1076

**Finding:** On current master this is already handled by `ExcludeBattleParticipantsRule` (Clash / You Are Beaten → may-not-battle → excluded → INACTIVE → no weapon top-level actions). The PR adds VHD regression tests only. Related #1072 (already-battled react / #297) is a different path and untouched.

PR #1072, opened 2026-09-12
Fix #297: reacted already-battled cards become INACTIVE in 2nd battle

### LINKED ISSUE #297: A atst from battle 1 reacted to battle 2 and was able to trample character in battle 2
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=advocate$ku5v76okh77kt5sj Turn 3 lightside

They missed thankfully.  But it should not be allowed to play that since it was inactive (from battling already that turn).

### LINKED ISSUE #554: applying gametext to a battle while excluded
## **Describe the bug**
I attacked at an adjected site that once my opponent reacts to the second battle at the adjected site his reacting char/vehicle will be excluded. This worked fine, however the gametext of the battled char still trickered (draw 2 if not able to otherwise when with an other droid). That other droid was excluded though and I think this doesnt count as present (but I could be wrong).

here is the link to the game: 

https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=dillen$w8jz9rr6zwmoqr1j

billbisco:
## Summary
Cards that move as a react into a second battle correctly may not participate, but stayed ACTIVE instead of becoming INACTIVE. `updateParticipants` only considered `canParticipateInBattleAt` candidates, so already-battled arrivals never received `ExcludedFromBattleModifier` and could still enable Trample, fire weapons, and apply gametext.

This applies the exclude modifier for present-but-ineligible cards at the battle location after the normal participant update, so they are inactive for the rest of the battle. React itself remains legal because the card is not marked inactive before it arrives.

Closes #297. Closes #554.

Does not claim #57 (Clash / exclude while already participating is a different path).

## Tip
`bb412a970332a99adea5b5ed361ead1295b2cb65`

## Test plan
- [x] `TrampleCannotTargetExcludedCharacterDuringBattle_Barrier` — Imperial Barrier exclude still inactive / not Trample-targetable
- [x] `TrampleCannotTargetExcludedCharacterDuringBattle_MultipleBattles` — reacted already-battled character is inactive / not Trample-targetable
- [x] `TrampleExcludedWalkerCannotTargetCharacterDuringBattle_MultipleBattles` — reacted already-battled AT-ST is inactive / cannot enable Trample
- [x] `Card_3_138_Tests` full class: 7 run / 0 fail / 0 skipped
- [x] Overlay **17004** rebuilt and healthy

### DIFF
--- a/src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/game/state/BattleState.java
+++ b/src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/game/state/BattleState.java
@@ -280,6 +280,30 @@ public class BattleState implements Snapshotable<BattleState> {
             }
         }
 
+        // Cards that arrive mid-battle but cannot participate (e.g. already battled elsewhere this turn
+        // via a react) never enter currentParticipants, so the exclude modifier above never applied.
+        // Apply ExcludedFromBattleModifier so getCardState returns INACTIVE for the rest of the battle.
+        // Do not mark them inactive before they arrive — react itself must still be legal.
+        Collection<PhysicalCard> presentButIneligible;
+        if (_isLocalTrouble) {
+            presentButIneligible = Filters.filterActive(game, null, Filters.and(
+                    Filters.initiallyParticipatesInBattle(_location),
+                    Filters.in(_localTroubleParticipants),
+                    Filters.not(Filters.in(currentParticipants)),
+                    Filters.not(Filters.in(getAllCardsParticipating()))));
+        }
+        else {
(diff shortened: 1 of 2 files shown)


billbisco:
**Tip** `bb412a970332a99adea5b5ed361ead1295b2cb65`

`Card_3_138_Tests`: **7 run, 0 fail, 0 error, 0 skipped** (includes Barrier exclude + both previously-ignored multi-battle react cases).

Files: `BattleState.java`, `Card_3_138_Tests.java`.

Closes #297; same path covers #554. Does not claim #57 (Clash/exclude-while-already-participating is a different path).

Overlay: **17004** only.

PR #1073, opened 2026-09-12
Fix Blaster Deflection LOST retarget for non-free weapons (#859)

### LINKED ISSUE #859: Blaster Deflection not giving option to use LOST
- [ ] Did you search through the list of **open issues** _first_ to ensure that there is not already an report for this issue? Yes

## **Describe the bug**
When a character with ability greater than 4 is armed with a lightsaber and targeted with a blaster, the card Blaster Deflection (LS) will only allow the USED version of the card to be played, not the LOST version. I tried with permanent weapon, matching saber, and generic saber all failed. LS had the required 3 force at the time of targeting.

## **To Reproduce**
Deploy a character with a blaster and a character ability 5 or higher with a lightsaber. Target the character with the lightsaber and try to play Blaster Deflection LOST

## **Game Replay ID:**

https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=WormTime$cs8eu50vir4eh4qf

## **Expected behavior**

Should present the player with options for both USED and LOST version of the card.

billbisco:
Why: leftover #859. Blaster Deflection LOST was not offered when retargeting a non-free blaster (Imperial Blaster). Deleting the affordable fire-target snapshot made LOST work, but fire targeting would then include cards that cost too much Force.

What it Does: Fire targeting still uses the affordable snapshot. acceptsIgnoringOwner (weaponMayRetargetTo) keeps the logical filter so LOST can retarget to the firer's characters. Card_6_061_Tests: stats, USED cancel, and LOST on non-free / free / repeated / built-in (EPP) blasters.

What it Fixes: Blaster Deflection LOST is offered and can retarget a non-free blaster. Fire targeting still only offers affordable targets.

Issues: Closes #859

### DIFF
--- a/src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/logic/actions/FireWeaponActionBuilder.java
+++ b/src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/logic/actions/FireWeaponActionBuilder.java
@@ -389,7 +389,10 @@ public class FireWeaponActionBuilder {
                             }
                         }
                     }
-                    _targetFilterList.set(i, Filters.in(validTargets));
+                    // Keep the logical target filter (same structure as free targeting).
+                    // Replacing it with Filters.in(validTargets) snapshots only currently
+                    // affordable opponent cards, which breaks Filters.weaponMayRetargetTo
+                    // (acceptsIgnoringOwner) used by retarget interrupts like Blaster Deflection LOST.
                     isValid = (validTargets.size() >= _numTargets);
                 }
                 _targetFilterValid.add(isValid);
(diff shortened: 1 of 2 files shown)


src/gemp-swccg-logic/src/main/java/com/gempukku/swccgo/logic/actions/FireWeaponActionBuilder.java
```diff
@@ -389,7 +389,10 @@ else if (_forFree || _targetsForFree.get(i)) {
                             }
                         }
                     }
-                    _targetFilterList.set(i, Filters.in(validTargets));
```
BOTVHD: I'm pretty suspicious of this being the correct/right fix.  I don't doubt that it allows Blaster Deflection LOST to work correctly, but I'd be surprised if it didn't break something else with weapon firing.

Still, knowing the specific part of code that is causing the issue is helpful.  And maybe I'm wrong and this line is unnecessary/incorrect in all cases...

billbisco: You're right. Dropping the affordable snapshot let LOST work, but fire targeting would then include cards that cost too much Force.

Fire targeting still uses the affordable snapshot. `acceptsIgnoringOwner` (used by `weaponMayRetargetTo`) now keeps the logical filter so Blaster Deflection LOST can retarget. Thanks VHD.

Issue #943, opened 2026-02-27
Stolen Restraining Bolt affects non-droid holder


Gergall:
- [x] Did you search through the list of **open issues** _first_ to ensure that there is not already an report for this issue?

## **Describe the bug**
Restraining Bolt affects anybody holding it, but it should only affect droids.

## **To Reproduce**
1. Give a character Merc Sunlet
2. Have them steal a Restraining Bolt
3. Character is affected even though they are not a droid

## **Expected behavior**
It should affect any droid (on either side) who tries to carry it, but not other characters.

## **Screenshots**
<img width="1610" height="1166" alt="Image" src="https://github.com/user-attachments/assets/69207898-e1c6-4d30-8f53-4a1aadfca146" />

## **Additional context**
Restraining Bolt is an unusual card because it can deploy on either player's droid.
The Scomp Rules page for Restraining Bolt has many rulings for intricate scenarios involving this card.
Apparently, the ability to transfer a Restraining Bolt is not working properly either, but not sure if we want to try to tackle that at the same time or separately.

BOTVHD:
An easy fix, I think, is to change this:
```
    @Override
    protected List<Modifier> getGameTextWhileActiveInPlayModifiers(SwccgGame game, final PhysicalCard self) {
        Filter hasAttached = Filters.hasAttached(self);

        List<Modifier> modifiers = new LinkedList<Modifier>();
        modifiers.add(new MayNotMoveModifier(self, hasAttached));
        modifiers.add(new CancelsGameTextModifier(self, hasAttached));
        return modifiers;
```
to this:
```
    @Override
    protected List<Modifier> getGameTextWhileActiveInPlayModifiers(SwccgGame game, final PhysicalCard self) {
        Filter hasAttached = Filters.and(Filters.hasAttached(self),Filters.droid);

        List<Modifier> modifiers = new LinkedList<Modifier>();
        modifiers.add(new MayNotMoveModifier(self, hasAttached));
        modifiers.add(new CancelsGameTextModifier(self, hasAttached));
        return modifiers;
```

However, it still makes sense to pursue the active/inactive angle, since it may have some additional implications.

This also means that if R2 uses Retractable Arm to steal a restraining bolt, it will shut off R2's gametext (basically stealing the Restraining Bolt and applying it to itself!) - I'm not sure if that is correct from a rules standpoint.  I'd guess yes, even though it was probably never intended to work quite that way.

Gergall:
[...]

PR #1077, opened 2026-09-12
Fix #886: unique permanent-pilot personas not location-present

### LINKED ISSUE #886: Permanent pilot personas incorrectly handled for gametext checks on 'present' (ex: Hidden Weapons and Boba Fett In Slave I)

## **Describe the bug**
By rules, unique persona permanent pilots (think Boba Fett in Slave 1, Death Star Assault Squadron, etc) are present in their respective ship / enclosed vehicle, but not present at the location ( that the ship / enclosed vehicle is present at!).

However, some cards in GEMP are coded in a way that (incorrectly) treats the persona as being at the location.

Example 1:
Boba Fett in Slave 1 (combo card) can use Hidden Weapons

Example 2:
Death Star Assault Squadron incorrectly removes Luke's +2 power bonus from Run Luke, Run!

Forum discussion: https://forum.starwarsccg.org/viewtopic.php?t=86010

## **To Reproduce**
1. Deploy (or land) Boba Fett in Slave 1 to a docking bay
2. Deploy a LS character to the docking bay
3. Have Hidden Weapons in DS hand
4. Battle at the docking bay
5. During weapons phase, observe that Hidden Weapons can be played (incorrectly!)

## **Expected behavior**
Permanent pilot personas should not be considered 'present' at sites/systems. In the example above, Hidden Weapons should not be playable.

## **Additional context**
I think this is happening because when checking if cards can be played, the card logic checks the persona filter (ex: Filters.Boba_Fett) and then does a simple 'and' operation with a present filter (ex: Filters.presentInBattle).  Both of these filters include the starship with unique permanent pilot card (ex: Boba Fett in Slave 1) and the logic incorrectly handles this as the persona being present.

Need to review other instances of named personas in ships/vehicles and where those personas are used in card gametext along with the term 'present'.

Some examples to check are:
- (landed) Death Star Assault Squadron (Vader as permanent pilot) and Your Destiny
- (landed) Death Star Assault Squadron (Vader as permanent pilot) and Physical Choke
- (landed) Han, Chewie, And The Falcon (Han as permanent pilot) and This Is Absolutely Right
- (landed) Han, Chewie, And The Falcon (Han as permanent pilot) and I've Got A Bad Feeling About This

Some cards are working OK (probably due to additional filter conditions?) like:
- (landed) Death Star Assault Squadron (Vader as permanent pilot) is (correctly!) unable to play The Circle Is Now Complete at the same site as Obi-Wan

billbisco:
## Summary
Unique permanent-pilot personas (Boba Fett In Slave I, Death Star Assault Squadron, Han/Chewie Falcon, etc.) are present aboard the craft, not at the location. Cards that AND a persona filter with `presentInBattle` / `presentAt` / `present` / `presentWith` incorrectly treated the combo ship as the persona being location-present.

Shared helpers `personaPresentInBattle` / `personaPresentAt` / `personaAt` match characters with the persona (and permanent pilots only aboard open vehicles). This PR wires the reported cards plus the prioritized Decipher sisters onto those helpers.

## Approach
Shared **personaPresent-style filters** in `Filters` (not card-local character-AND hacks alone). Open-vehicle permanent pilots remain location-present; starships and enclosed vehicles do not.

## Cards updated this PR
- Hidden Weapons (`6_154`)
- Run Luke, Run! (`101_3`)
- Your Destiny (`9_134`)
- Physical Choke (`1_264`)
- Focused Attack (`5_141`)
- I've Got A Bad Feeling About This (`1_91`)
- This Is Absolutely Right (`8_62`)
- I Will Find Them Quickly, Master (`11_71`)
- Anakin Skywalker (`9_48`)
- Field Promotion (`4_121`)

Skipped mass Virtual-only rewrites. Left for rules (not just filter wiring): Maul Strikes "in battle" vs "present with" duel branch, and other presentWith/in-battle edge cases.

## Tip
`3684ae2a356fba74de569b683ea50af143eef347`

## Tests (overlay 17004)
Prioritized suites — **12 run / 0 fail / 0 skip**
- `Card_6_154_Tests` — 2
- `Card_101_003_Tests` — 1
- `Card_9_134_Tests` — 2 (DSAS vs character Vader / Your Destiny)
- `Card_5_141_Tests` — 2 (Focused Attack + DSAS)
- `Card_1_091_Tests` — 2 (Falcon vs character Han / Bad Feeling)
- `Card_8_062_Tests` — 2 (Falcon vs Han+Leia / Absolutely Right)
- `Card_1_264_Tests` — 1 (Physical Choke Rebel Trooper path; Dark_Jedi already character-gated)

## Overlay
Overlay **17004** only.

Closes #886

Skipped #943 (device active/inactive).

### DIFF
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set1/dark/Card1_264.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set1/dark/Card1_264.java
@@ -89,7 +89,7 @@ public class Card1_264 extends AbstractLostInterrupt {
         // Check condition(s)
         if (TriggerConditions.wonBattle(game, effectResult, playerId)
                 && GameConditions.canUseForceToPlayInterrupt(game, playerId, self, 1)
-                && GameConditions.isDuringBattleWithParticipant(game, Filters.and(Filters.Dark_Jedi, Filters.presentInBattle))
+                && GameConditions.isDuringBattleWithParticipant(game, Filters.personaPresentInBattle(Filters.Dark_Jedi))
                 && GameConditions.canTarget(game, self, targetingReason, filter)) {
 
             final PlayInterruptAction action = new PlayInterruptAction(game, self);
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set1/light/Card1_091.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set1/light/Card1_091.java
@@ -44,7 +44,7 @@ public class Card1_091 extends AbstractUsedInterrupt {
         // Check condition(s)
         if (TriggerConditions.battleInitiatedAt(game, effectResult, playerId, Filters.canTargetCard(self))
                 && GameConditions.hasLessPowerInBattleThanOpponent(game, playerId)) {
-            final int damageMultiplier = GameConditions.isDuringBattleWithParticipant(game, Filters.and(Filters.Han, Filters.presentInBattle, Filters.canBeTargetedBy(self))) ? 3 : 2;
+            final int damageMultiplier = GameConditions.isDuringBattleWithParticipant(game, Filters.and(Filters.personaPresentInBattle(Filters.Han), Filters.canBeTargetedBy(self))) ? 3 : 2;
 
             final PlayInterruptAction action = new PlayInterruptAction(game, self);
             action.setText(damageMultiplier == 2 ? "Double opponent's battle damage" : "Triple opponent's battle damage");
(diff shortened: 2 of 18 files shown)


src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set1/dark/Card1_264.java
BOTVHD, line 1: I actually don't think any change is needed to this card (Dark Jedi is not a persona) and it is probably working correctly already.

BOTVHD:
If we went this route, it might make more sense to pass in a persona instead of a filter?

For example:
personaPresentInBattle(Persona.BOBA_FETT)

rather than:
personaPresentInBattle(Filters.Boba_Fett)

Also, I'm not sure, but it looks like this may only work for characters.  I'm guessing personaPresentInBattle called on the Falcon would return false even if the Falcon was present at the battle.

Not suggesting any changes right now, though.

billbisco:
Agreed on Physical Choke. Dark Jedi is not a persona, so that check is back to `Filters.Dark_Jedi` + `presentInBattle`.

On the helper taking `Persona` vs `Filter`, and Falcon/starship coverage: leaving that for a later pass as you said no changes needed there right now.

Issue #998, opened 2026-08-24
Endor Dark Forest not recognizing Wuta (v)


hayesbug:
- [ YES ] Did you search through the list of **open issues** _first_ to ensure that there is not already an report for this issue?

## **Describe the bug**
Wuta (v) was present at Dark Side's Endor: Dark Forest site ('If your Ewok present, Force drain +1 here.")
When Light Side drained on Turn 3, no bonus was added - it was just a drain of 2

## **To Reproduce**
See replay

## **Game Replay ID:**
Light side Turn 3
[(https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=Hayes$e0iub9f47ha4vohi)

## **Expected behavior**
Expected drain to be for 3 - it was only a drain of 2

## **Screenshots**
n/a

## Client Platform
Google Chrome

BOTVHD:
Hmmm, I'm stumped on a cause here.

I don't see any cards played that would change the drain from being 2 + 1.
The cards immediately involved look like they are written correctly:
- Wuta (V) has the Ewok species property
- Dark Forest has the correct 'your Ewok present' condition set for a drain +1 modifier

I tried a simple test game with Wuta (V) and the other characters at Dark Forest and drained for 3 correctly.

Possible this is a lost modifier #911 case on the Dark Forest's drain +1 modifier.

~~You didn't happen to right click on Dark Forest to see if it showed the drain +1 modifier (and/or any other modifiers), did you?~~

EDIT: Confirmed that during game, right clicking on Wuta (V) showed Ewok species but the Dark Forest did show the +1 drain modifier.

billbisco:
Added tests that match what VHD saw in a simple game: LS Force drain at Dark Forest with original Wuta or Wuta (V) present is 3 (2 Dark icons + 1 Ewok).

PC: https://github.com/PlayersCommittee/gemp-swccg-public/pull/1091

That means #998 is not “Wuta (V) is not an Ewok.” If the replay still drained 2 while the location showed +1, that lines up with VHD’s lost-modifier suspicion (#911). Next step is to reproduce a drain that ignores a modifier that is still listed on the location.

billbisco:
Open PR already covers this: https://github.com/PlayersCommittee/gemp-swccg-public/pull/1091

Printed Endor: Dark Forest drain with Wuta (V).

Issue #291, opened 2021-01-29
[Bug] Jodo Kast game text is a may action and is being enforced automatically.


DingusMyDingus:
So this has happened in multiple games using multiple weapons (crimson dawn blaster, stun blaster) at multiple locations and with different targets.  Game text of Jodo: "Once per turn when firing a rifle or blaster, may fire for free and add 2 to total weapon destiny".  However the first firing each turn is enforcing this 'for free' and +2.

This is especially problematic when using defensive fire (v) as I would forgo his gametext on the already 'free' shot that defensive fire (v) affords and then use it when firing again normally in battle.

Same issue when using Sniper.

BOTVHD:
Note that FireWeaponFiredByForFreeModifier is only used on Jodo Kast and Sabine Wren - and both of them have the same 'may' in their game text.

BOTVHD:
This probably gets tricky because of the cost.

Deciding if the weapon action builder should provide the action depends on if you can pay the force cost.  If you have 0 force available and don't have the 'free' modifier already applied, then the weapon action wouldn't be available to choose.

I'm not sure if there is a good way to make a parallel free version of the firing action to check for (so you can pick which one to use, if your once per turn free shot hasn't been used yet).

Medium Repeating Blaster Cannon builds two separate firing actions, but I'm not sure if Jodo could bestow an optional FireWeaponAction that duplicates whatever is in the getGameTextFireWeaponActions (but replaces the cost with 'free').

Arica's fire weapon text may be worth looking at, too (but not sure if this would fully resolve things like use with Sniper).

billbisco:
This is implemented by #1085 and when complete will close this issue.

PR #1083, opened 2026-09-12
Fix draw-X-choose-Y when max battle destinies already reduced (#973)

### LINKED ISSUE #973: Droid starfighter draw 2 choose 1
I asked.

## **Describe the bug**

https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=advocate$irt3clh29qewk2g2

Turn 3, battle at Yavin IV.  He had the flagship with a b2v droid aboard and the DFS1308 and DFS1015 at the system when I battled him with two ships and haven.  

He was only able to draw one destiny (because of the b2 if unable to text) and didn't have 4 ability.  But the draw two text and choose one should still work (from the 1015 droid) but it didn't work.  Bug?

## **To Reproduce**
3. 
4.
5. See error

billbisco:
## Summary
When max battle destinies is already reduced (e.g. to 1), draw-X-choose-Y text such as DFS-1015's "draw two and choose one" was incorrectly blocked because `canDrawAndChoose` compared **draw-X** against the max instead of **choose-Y**.

## Fix
- `DrawDestinyEffect.canDrawAndChoose` / `GameConditions.canDrawDestinyAndChoose` now gate on **choose-Y** (destinies that count toward the limit).
- Mid-draw limit handling allows finishing the remaining draws within X; only the first choose-Y draws increment `_numDrawnSoFarAgainstLimit`.
- Decipher call sites updated to pass choose-Y: DFS-1015, Take This!, Gambler's Luck, Gray Squadron Y-wing Pilot.
- `DrawDestinyAndChooseInsteadEffect` validates with choose-Y.

## Tests
- `Card_14_116_Tests` (3): DFS-1015 stats; draw-2-choose-1 offered when max=1; choose-Y=2 not offered when max=1.
- `Card_5_048_Tests` (1): Gambler's Luck stats (call-site choose-Y update).

## Independent
Independent PR vs `master`. Does not include #291 / #998 work.

## Known gaps
- **Reviewed / deferred (Alpha2):** against-limit if more battle destinies remain in the same `DrawDestinyEffect` after a draw-and-choose. Primary real-path check (DFS-1015 draw-2-choose-1 with max/count=2, then the second normal destiny in the same effect) already completes correctly under this PR's choose-Y against-limit accounting (`numDrawn=2`). Not expanding this PR with extra coverage. Residual follow-up only if exotic mid-draw-X cancel/substitute/skip accounting proves wrong.
- Replay used Virtual DFS / B2 as incidental setup; tests use Decipher DFS-1015 + ad-hoc max/min destiny modifiers.

Closes #973

### DIFF
--- a/src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set14/dark/Card_14_116_Tests.java
+++ b/src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set14/dark/Card_14_116_Tests.java
@@ -1,0 +1,198 @@
+package com.gempukku.swccgo.cards.set14.dark;
+
+import com.gempukku.swccgo.common.CardSubtype;
+import com.gempukku.swccgo.common.CardType;
+import com.gempukku.swccgo.common.ExpansionSet;
+import com.gempukku.swccgo.common.Icon;
+import com.gempukku.swccgo.common.Keyword;
+import com.gempukku.swccgo.common.ModelType;
+import com.gempukku.swccgo.common.Phase;
+import com.gempukku.swccgo.common.Rarity;
+import com.gempukku.swccgo.common.Side;
+import com.gempukku.swccgo.common.Uniqueness;
+import com.gempukku.swccgo.cards.GameConditions;
+import com.gempukku.swccgo.framework.StartingSetup;
+import com.gempukku.swccgo.game.AbstractActionProxy;
+import com.gempukku.swccgo.game.PhysicalCard;
+import com.gempukku.swccgo.game.SwccgGame;
+import com.gempukku.swccgo.logic.TriggerConditions;
+import com.gempukku.swccgo.logic.actions.OptionalGameTextTriggerAction;
+import com.gempukku.swccgo.logic.actions.TriggerAction;
+import com.gempukku.swccgo.logic.timing.EffectResult;
+import java.util.Collections;
+import java.util.List;
+import com.gempukku.swccgo.framework.VirtualTableScenario;
+import com.gempukku.swccgo.logic.decisions.AwaitingDecision;
+import com.gempukku.swccgo.logic.modifiers.DrawsBattleDestinyIfUnableToOtherwiseModifier;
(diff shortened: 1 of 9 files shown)


billbisco:
**Known gap clarification (GAP_DEFER):** Follow-up diagnostic on tip `4dd37f5` found the shipped choose-Y against-limit already allows finishing peeks within X while only counting Y against the limit. A real-path check (DFS-1015 draw-2-choose-1 with max/count = 2, then a second normal destiny) completed with `numDrawn=2` — no failing “more destinies remain after draw-and-choose” hole in this PR’s scope.

Residual exotic edges (cancel / substitute / skip mid-draw-X) are out of scope here and may need a separate issue if they appear in play.

src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set14/dark/Card_14_116_Tests.java
```diff
@@ -0,0 +1,198 @@
...
+
+        var xwing = scn.GetLSCard("xwing");
+        var dfs1015 = scn.GetDSCard("dfs1015");
+        var dfs1308 = scn.GetDSCard("dfs1308");
+        var system = scn.GetDSStartingLocation();
+
+        scn.StartGame();
+        scn.MoveCardsToLocation(system, dfs1015, dfs1308, xwing);
+
+        scn.ApplyAdHocModifier(new DrawsBattleDestinyIfUnableToOtherwiseModifier(dfs1015, 1));
```
BOTVHD, line 156: Adding all these "AdHoc" modifiers and actions is a useful way to develop the card changes/fixes, but I'm not convinced it is a good thing for long term tests.

I think a set of tests with specific cards in Reserve and playing out a "draw X choose Y" card (with various responses, like substituting, cancelling, redrawing) destinies would be better.

billbisco:
Thank you, VHD. The max-destiny cap in the DFS-1015 tests is now Captain Madakor piloting Radiant VII (opponent may not draw more than one battle destiny here) instead of an AdHoc modifier. DFS-1015 itself is still the draw-two-choose-one card under that printed cap.

PR #1084, opened 2026-09-12
Fix #943: stolen devices inactive for invalid holders

### LINKED ISSUE #943: Stolen Restraining Bolt affects non-droid holder

billbisco:
## Summary
Stolen character devices were still applying while-active-in-play modifiers when held by someone who is not a valid user of the device (e.g. Restraining Bolt on a non-droid; Stormtrooper Utility Belt on a Rebel). Character weapons already become inactive in that situation; devices did not.

Root cause: `AbstractCharacterDevice` never set `CardSubtype.CHARACTER`, so the existing check in `Cards.getCardState` (which gates on weapon/device + CHARACTER subtype + `getValidToUseDeviceFilter`) never ran for devices.

## Fix
Set `CardSubtype.CHARACTER` on `AbstractCharacterDevice`, mirroring `AbstractCharacterWeapon`. Inactive state then correctly follows `getGameTextValidToUseDeviceFilter` — no bolt-only `Filters.droid` workaround and no per-card modifier rewrites.

## Tests
VHD (`Card_1_205_Tests`, 5/5):
- Restraining Bolt reattached from a droid onto a non-droid → bolt inactive; no may-not-move on the non-droid
- Restraining Bolt on a droid still applies may-not-move
- Stormtrooper Utility Belt reattached onto a Rebel → belt inactive; no +1 power
- Vibro-Ax on an invalid user still does not give +1 (weapons path unchanged)
- Stats assert includes Character subtype

Closes #943

### DIFF
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/AbstractCharacterDevice.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/AbstractCharacterDevice.java
@@ -1,5 +1,6 @@
 package com.gempukku.swccgo.cards;
 
+import com.gempukku.swccgo.common.CardSubtype;
 import com.gempukku.swccgo.common.ExpansionSet;
 import com.gempukku.swccgo.common.Keyword;
 import com.gempukku.swccgo.common.PlayCardZoneOption;
@@ -25,10 +26,11 @@ public abstract class AbstractCharacterDevice extends AbstractDevice {
      * @param title the card title
      * @param uniqueness the uniqueness
      * @param expansionSet the expansionSet
-￼    * @param rarity the rarity
-￼    */
+     * @param rarity the rarity
+     */
     public AbstractCharacterDevice(Side side, float destiny, String title, Uniqueness uniqueness, ExpansionSet expansionSet, Rarity rarity) {
         super(side, destiny, PlayCardZoneOption.ATTACHED, title, uniqueness, expansionSet, rarity);
+        setCardSubtype(CardSubtype.CHARACTER);
         addKeywords(Keyword.DEPLOYS_ON_CHARACTERS);
     }
 
--- a/src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set1/dark/Card_1_205_Tests.java
+++ b/src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set1/dark/Card_1_205_Tests.java
@@ -2,8 +2,6 @@ package com.gempukku.swccgo.cards.set1.dark;
 
 import com.gempukku.swccgo.common.CardSubtype;
 import com.gempukku.swccgo.common.CardType;
-import com.gempukku.swccgo.common.Phase;
-import com.gempukku.swccgo.common.PlayCardOptionId;
 import com.gempukku.swccgo.common.Side;
 import com.gempukku.swccgo.common.Uniqueness;
 import com.gempukku.swccgo.framework.StartingSetup;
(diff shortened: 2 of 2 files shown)


BOTVHD:
Technically, there is no subtype of "character" for Devices (like this for weapons):
image

They are all just "Device" card type.
image

I don't think we should use the subtype approach (especially because there could be other usages looking at subtypes?).

Does it work just to change Cards.java line 67-69 from:
```
            if (cardCategory == CardCategory.DEVICE && cardSubtype == CardSubtype.CHARACTER && physicalCard.getAttachedTo() != null
                    && !physicalCard.getBlueprint().getValidToUseDeviceFilter(physicalCard.getOwner(), gameState.getGame(), physicalCard).accepts(gameState, query(), physicalCard.getAttachedTo()))
                return CardState.INACTIVE;
```

to:
```
            if (cardCategory == CardCategory.DEVICE && physicalCard.getAttachedTo() != null && physicalCard.getAttachedTo().getBlueprint().getCardCategory() == CardCategory.CHARACTER
                    && !physicalCard.getBlueprint().getValidToUseDeviceFilter(physicalCard.getOwner(), gameState.getGame(), physicalCard).accepts(gameState, query(), physicalCard.getAttachedTo()))
                return CardState.INACTIVE;
```

BOTVHD:
Also, unrelated, but it seems like 'includeWeaponsForStealingForThisCard' should be renamed 'includeWeaponsAndDevicesForStealingForThisCard' just to be clearer.

billbisco:
Agreed. Devices are not a character subtype, so invalid-holder inactivity now checks that the host is a character. Stealing also includes devices (not only character-subtype weapons). Renamed the local steal flag to `includeWeaponsAndDevicesForStealingForThisCard`.

src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/AbstractCharacterDevice.java
```diff
@@ -25,10 +26,11 @@ public abstract class AbstractCharacterDevice extends AbstractDevice {
      * @param title the card title
      * @param uniqueness the uniqueness
      * @param expansionSet the expansionSet
-￼    * @param rarity the rarity
-￼    */
+     * @param rarity the rarity
+     */
     public AbstractCharacterDevice(Side side, float destiny, String title, Uniqueness uniqueness, ExpansionSet expansionSet, Rarity rarity) {
         super(side, destiny, PlayCardZoneOption.ATTACHED, title, uniqueness, expansionSet, rarity);
+        setCardSubtype(CardSubtype.CHARACTER);
```
BOTVHD, line 33: This is bad - there is no character subtype for Device cards.

billbisco:
Thank you, VHD.

Agreed: devices have no character subtype. Removed setCardSubtype(CHARACTER) from AbstractCharacterDevice. Invalid-holder inactivity is only the Cards.getCardState check that the host is a character who fails getValidToUseDeviceFilter. Stats no longer assert a character subtype.

PR #1085, opened 2026-09-12
Prompt Jodo Kast / Sabine Wren May when firing a weapon (#291)

billbisco:
https://github.com/PlayersCommittee/gemp-swccg-public/issues/291 One Fire action, then Yes/No to fire for free and add 2. Same prompt on Sniper, Defensive Fire (V), and Sorry About The Mess. Defensive Fire each-destiny +2 stacks with Jodo total +2.


BOTVHD:
Probably need to get rules clarification about this question:

> if you take this 'may' action, is it required that you use both the 'free' and 'add 2' components?

billbisco:
@BOTVHD Confirmed: targeting for free and adding 2 are **one Action**, not two independent mays.

If you take the once-per-turn, you get both (free target **and** +2 total weapon destiny). You cannot take free fire without the +2, or the +2 without free fire.

The PR already presents that as a single weapons-segment action (`Fire … for free and add 2`) alongside the normal paid fire. Paid fire does not consume the package. Using the package consumes it for the turn.

That is the ruling we are shipping.

billbisco:
Replaced the two separate Fire buttons with a Yes/No prompt on every fire: `Use Jodo Kast: fire for free and add 2 to total weapon destiny?` Same for Sabine. Tests cover battle, Sniper, Defensive Fire (V) (skip May, and stack +2/+2), and Sorry About The Mess.

PR #1010, opened 2026-09-01
Fix Cumulative Rule in the modifier engine (issues 38 and 697)

billbisco:
Fixes Cumulative Rule in the modifier engine so non-cumulative modifiers do not stack unless they say Cumulatively.

## Summary
Engine-level Cumulative Rule for issues #38 and #697. Not a Concentrate All Fire title special-case — the hole was unique same-`cardId` re-emission of the same `ModifierType`, plus same-title copies and same game-text clause from two cards.

## Card / scope
- Engine change (`foundCumulativeConflict` / clause identity), with card tags only where needed
- Touches Concentrate All Fire (`9_003`), Fighter Cover (`9_093`), Rebel Flight Suit (`4_014`), Mercenary Pilot (`112_013`) clause tags, Darth Maul Young Apprentice (`12_101`), Sai'torr Kal Fas (`1_64`), Jawa Sandcrawler (`1_309`), Rebel Tech (`2_019`)

## What changed
Conflict when a new modifier matches an already-kept one on: not cumulative, has source, same `ModifierType`, same permanent-pilot/astromech flags, same source title, same `forPlayer`/icon, and same **clause identity**.

Clause identity API on `Modifier` / `AbstractModifier`:
- `setCumulative` / `getClauseIdentity` / `setClauseIdentity` / `applyClauseIdentityFromAction`
- Auto-tag from `AddModifierWithDurationEffect`, `ModifyPowerUntilEndOfBattleEffect`, `ModifyPowerUntilEndOfTurnEffect`

Card tags:
- **Mercenary Pilot** (`112_013`): driving → `OTHER_CARD_ACTION_1`; cloud-sector optional → `OTHER_CARD_ACTION_2` (optional behavior itself is #1011)
- **Rebel Flight Suit** (`4_014`): removed fake `cumulative=true`; clause so matching-pilot maneuver combines with pilot native +1 then limit +2

## Design choices
- Different clause identities do not conflict (issue 697)
- `isCumulative()` always stacks (Rebel Tech)
- Power summing still adds every surviving POWER modifier; de-dupe is only at conflict check
- Against master (not stacked on #1006 / #1007 / #1011); #1011 is mergeable first

## Tests
[...]

PR #1088, opened 2026-09-12
Fix Concentrate All Fire stacking on same starfighter (#38)

### LINKED ISSUE #38: Concentrate All Fire + B-Wings

### LINKED ISSUE #46: Fighter Cover Accumulation (TIE INTERCEPTOR)

billbisco:
## Summary
Closes #38.
Closes #46.

Concentrate All Fire was adding +3 power every time a starfighter fired a weapon in the same battle (B-Wings with multiple weapons). Per ruling, each starfighter should get that bonus only once per battle. Fighter Cover (Decipher twin) shares the same trigger and gate (#46).

## Approach
Minimal card-level once-per-target-per-battle gate (no cumulative-engine rewrite):
- Before applying the until-end-of-battle +3, skip if that starfighter already has a POWER modifier sourced from this Admirals Order.
- Same gate on Fighter Cover (Decipher twin with the same trigger).
- Maul Strikes / broader cumulative-rule work left soft-hold.

## Tests
`Card_9_003_Tests` — **2 run / 0 fail / 0 skip**
- `ConcentrateAllFireStatsAndKeywordsAreCorrect`
- `ConcentrateAllFireDoesNotAddPowerTwiceToSameStarfighterInOneBattle` (multi-weapon B-Wing: first fire +3, second fire not offered/not applied; power stays 7)

`Card_9_093_Tests` — **2 run / 0 fail / 0 skip**
- `FighterCoverStatsAndKeywordsAreCorrect`
- `FighterCoverDoesNotAddPowerTwiceToSameStarfighterInOneBattle` (twin path: DS Fighter Cover + LS multi-weapon B-Wing; same once-per-starfighter gate)

Combined surefire: **4 run / 0 fail / 0 skip**

## Tip
`024e75b3bbf35d9631a5cfd6401bd0bb339fe33b`

### DIFF
--- a/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set9/dark/Card9_093.java
+++ b/src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set9/dark/Card9_093.java
@@ -17,6 +17,7 @@ import com.gempukku.swccgo.logic.effects.ModifyPowerUntilEndOfBattleEffect;
 import com.gempukku.swccgo.logic.modifiers.EachWeaponDestinyModifier;
 import com.gempukku.swccgo.logic.modifiers.ImmunityToAttritionChangeModifier;
 import com.gempukku.swccgo.logic.modifiers.Modifier;
+import com.gempukku.swccgo.logic.modifiers.ModifierType;
 import com.gempukku.swccgo.logic.timing.EffectResult;
 import com.gempukku.swccgo.logic.timing.results.FiredWeaponResult;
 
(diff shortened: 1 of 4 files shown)


BOTVHD:
This is a workaround for an incomplete cumulative rule framework.  If/when something like this gets done (as triage until the cumulative rule is fixed), we definitely would want to make some kind of 'todo' list and keep track of these workarounds so they can be stripped out, eventually.

billbisco:
Withdrawing this PR.

https://github.com/PlayersCommittee/gemp-swccg-public/pull/1010 already fixes the same bugs at the modifier-engine level (issue #38 Concentrate All Fire, and Fighter Cover including the TIE two-weapon case from issue #46). This card-level once-per-starfighter gate is unnecessary and should not be implemented.

Closing in favor of #1010.

Issue #911, opened 2025-12-28
Lost modifier / AOBS bug


BOTVHD:
- [X] Did you search through the list of **open issues** _first_ to ensure that there is not already an report for this issue?

## **Describe the bug**
There has been a long-running bug, often called a 'lost modifier' bug or 'AOBS' bug where, during a game, part of a card's functionality (a modifier) stops working.

This is hopefully a place where we can collect information that we know about the issue and information we don't know (yet).  We can (hopefully) narrow the sources, perhaps add some diagnostic code, and eventually solve the issue.

For a list of (almost certain) instances of this bug, check the issue label "Lost modifier bug": https://github.com/PlayersCommittee/gemp-swccg-public/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22Lost%20modifier%20bug%22

## **To Reproduce**
Unclear how to reproduce.  Finding a way to reproduce this will probably make the solution relatively simple.

## **Game Replay ID:**
See above for individual issues with screenshots and replays of individual examples.

## **Assumptions**
This bug has been happening for years with no obvious dramatic increase/decrease in reporting of cases.  We can assume it is still a problem and that code changes over the years have not made a significant impact.

We assume this is a single underlying issue - but it's possible there are multiple root causes.

## **Observations**
1. **Likely only a small fraction of cases are being reported.**  Very noticeable bugs (like Presence Of The Force no longer providing modifiers for icons) are being reported because they impact gameplay, but a large portion of modifiers applied by cards never impact gameplay and if they stop working, players probably wouldn't even notice.
2. **Some reports that reverting can (but does not always) fix the modifier.** Unconfirmed? Can we get a replay of a case where reverting has 'fixed' a lost modifier?
3. **Presence Of The Force is a particularly valuable.** Watching replays, we can see the moment Presence of the Force modifiers stop working because the icon count (for player activation) changes on the UI.  For most of the other modifiers, we don't know when they occurred - we just see when the lack of that modifier first affects gameplay.
4. **Instance of losing modifiers from multiple copies of a card simultaneously.** #874 and #323 show cases where 2 active copies of Presence Of The Force both stop applying their modifiers.  It is not clear if this is always the case (all copies of a card lose their modifiers simultaneously). Do we have any replays where not all active copies are affected?
5. **Player / Turn timing.** Seen on DS and LS turns
6. **Turn phase timing.**
- After DS Battle Phase but before DS Move Phase. #874 
[...]

PR #1091, opened 2026-09-15
Test Endor Dark Forest Force drain with Wuta (V) (#998)

billbisco:
https://github.com/PlayersCommittee/gemp-swccg-public/issues/998 Tests that Dark Forest LS Force drain +1 counts original Wuta and Wuta (V).


billbisco:
The tests on this PR match a simple Dark Forest drain with original Wuta or Wuta (V): drain is 3.

If a live game still drained 2 while the location showed the +1 icon, that is the lost-modifier class (https://github.com/PlayersCommittee/gemp-swccg-public/issues/911), not Wuta (V) failing to count as an Ewok. Keep this open with 911.

Issue #939, opened 2026-02-20
Hobbie in Red Squadron 4 --> Cannot make X=3 with X-wing Laser Cannons


Icy13:
## **Describe the bug**
When Hobbie is Piloting Red Squadron 4, he may use 2 force to make X=3 on X-Wing Laser Cannons. GEMP correctly limited it to 2 force to spend but then applied X=2 rather than X=3. This caused a miss when it should have been a hit

## **To Reproduce**
1. Deploy Hobbie Piloting Red Suqadron 4 and Add X-Wing Laser Cannons
2. Initiate Battle
3. Target a Starfighter
4. Popup asks "Value for X and force to spend"
5. Max spend is 2 (which should create an X of 3, not an X of 2)

## **Game Replay ID:**
[No Replay yet (game in progress). Will respond with link when finsihed
](https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=Romaggedon$c5yh8utxpb1xqwum)

## **Expected behavior**
The Force cost should be 2, but X should be 3 (+3 to destiny)

## **Screenshots**
n/a

### **Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 -  - Version Version 144.0.7559.135 (Official Build) (64-bit)

bibfortuna25:
In order to use Red Squadron 4’s text, you need to select 0 for the initial value of X. Then it will prompt you if you want to use 2 to set X to 3.

billbisco:
The X prompt now lists a labeled choice: `X=3 using Red Squadron 4 (use 2 Force)`. Same for Saber 1 / SFS L-s9.3 at 1 Force. You can still pick a normal X and then use the ship text afterwards. PRs: https://github.com/billbisco/swccg-gemp/pull/95 and https://github.com/PlayersCommittee/gemp-swccg-public/pull/1095

PR #1095, opened 2026-09-15
Red Squadron 4 can set X-wing Laser Cannon X=3 for 2 Force (#939)

billbisco:
https://github.com/PlayersCommittee/gemp-swccg-public/issues/939 Firing prompt includes X=3 using Red Squadron 4 (use 2 Force). Same pattern for Saber 1.


billbisco:
Sorry — this PR was opened from the wrong base, so GitHub counted hundreds of unrelated commits against PC master. Rebuilt on PC master. It is now only the Red Squadron 4 X=3 firing prompt (and the one test that fires that prompt). Stats/icon tests are gone.

PR #1098, opened 2026-09-15
Captive Fury captives may forfeit when they are the only attrition payment (#951)

billbisco:
Why: leftover #951. Captive Fury was not allowing Light to forfeit a captive to pay remaining attrition.

What it Does: Non-hit captives are forfeitable only when attrition remains and they are the only remaining payment.

What it Fixes: When the captive is the only attrition payment left, Captive Fury lets (and requires) that forfeit.

Issues: Closes #951


BOTVHD:
At a glance, this seems like a good fix, but probably needs to be tested to see if it is working correctly with ignorable attrition.  My understanding from the rules clarification in the original issue is that if LSJK was battling with captive fury and LS has 1 (ignorable) attrition remaining (and no other targets eligible to forfeit), then LS should not be able to forfeit LSJK.

I think maybe in this change, it would (incorrectly) be allowed because the attritionRemaining condition would be satisfied.  It probably needs to change to a "no mandatory attrition remaining" and then the final modifier can be:
```
modifiers.add(new MayNotBeForfeitedInBattleModifier(self, Filters.and(Filters.in(chosenCaptives), Filters.not(Filters.hit)),
                        new OrCondition(otherForfeitableExists, noMandatoryAttritionRemaining)));
```

billbisco:
Thank you, VHD. Remaining attrition that Light is immune to is not a must-forfeit, so a non-hit captive still may not be forfeited.

The modifier now uses mandatory attrition (at least one present participating card is not immune to the total) instead of any remaining attrition. Premiere Luke immune to attrition < 3 is not offered at remaining 1; Chewie with no immunity still is when he is the only payment.

PR #1100, opened 2026-09-15
Squadron replacement transfer is optional (#965)

### LINKED ISSUE #965: Squadron replacement requires transferring all cards

## **Describe the bug**
Currently, GEMP requires the user to transfer all cards (pilots, passengers, devices, etc) from the starships to the squadron replacing those starships.  Pressing 'Done' after transferring some (but not all) cards does nothing.

According to the AR:

> The three 'replaced' starfighters and 
> all cards deployed on them go to their owner's Used 
> Pile. However, the **owner may choose to transfer any** 
> **characters, weapons and devices** to the squadron if 
> there is capacity (and the card may legally deploy on 
> squadrons); this transfer is free. Any cards targeting a 
> replaced starship transfer that targeting to the 
> squadron, unless they are not applicable (for example, 
> they only target starfighters) in which case they are lost.

 It should be possible, for example, to not transfer pilots aboard and those pilots would be placed in the Used Pile, along with the starfighters.

## **To Reproduce**
1. Have 3 Y-Wings with at least 2 other cards attached/aboard (ex: 2 pilots)
2. Begin a squadron replacement with Y-Wing Assault Squadron from hand
3. Transfer 1 card
4. Try to press 'Done' (leaving 1+ card attached/aboard)
5. Notice you are required to keep transferring until there are no more eligible cards to transfer.

## **Expected behavior**
Transferring should be optional (any, all, or none of the attached/aboard cards).

## Client Platform
N/A

billbisco:
Why: leftover #965. Squadron replacement was requiring every attached card to transfer. Printed replacement lets remaining cards go with the replaced starfighters.

What it Does: Done stops optional transfers. Remaining cards go to Used Pile with the replaced starfighters.

What it Fixes: Squadron replacement no longer forces transferring all cards.

Issues: Closes #965


BOTVHD:
This seems like a reasonable solution.  I would probably call the new variable selectedACardToTransfer instead of transferredOne, but that's largely personal preference.

billbisco:
Thank you, VHD.

Renamed transferredOne to selectedACardToTransfer.

Issue #999, opened 2026-08-25
Chewie's Bowcaster costs 3 Force to fire late in game


wjkafer:
- [ ] Did you search through the list of **open issues** _first_ to ensure that there is not already an report for this issue?

## **Describe the bug**
Chewie's Bowcaster is deployed on Chewbacca of Kashyyyk v. On LS turn 3 it fires normally. On LS turn 6 it costs 3 Force to fire.

## **To Reproduce**
3. 
4.
5. See error

## **Game Replay ID:**
[](https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=TacoBill$c63dcch5ipzik9df)

## **Screenshots**

<img width="416" height="84" alt="Image" src="https://github.com/user-attachments/assets/473cc3b0-aeda-4f7c-90cd-cdd9247d3514" />

<img width="407" height="57" alt="Image" src="https://github.com/user-attachments/assets/7482db27-52a3-4ddb-a756-008abe1a0991" />

BOTVHD:
If possible, it would be great to have a replay link.

As mentioned on slack earlier, it seems likely that this is a 'lost modifier' bug: #911

wjkafer:
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=TacoBill$c63dcch5ipzik9df

I could have sworn I added to the original, but oh well.

billbisco:
Looked at this as a possible #911 lost-modifier case.

**Printed text** (Chewbacca's Bowcaster): "May target a character, creature or vehicle using 3 Force (for free if Chewie firing)."

If Chewie is still the character firing, the cost is 0. A 3 Force cost later in the same game would mean the free-fire modifier stopped applying.

**What we verified:** with Chewbacca Of Kashyyyk and the bowcaster attached, Fire is still offered at 0 Force in the weapons segment. Tests: https://github.com/PlayersCommittee/gemp-swccg-public/pull/1101

**What we could not reproduce:** the late-game jump to 3 Force. Suggestion: keep this open as a #911 instance unless a replay shows Chewie was not the firer (then 3 Force is legal).

billbisco:
Watched the replay XML for [TacoBill$c63dcch5ipzik9df](https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=TacoBill$c63dcch5ipzik9df). The report is correct: this is a live #911 lost-modifier case, not a rules miss.

[...]

PR #1101, opened 2026-09-15
Test Chewbacca's Bowcaster fires for free on Chewie (#999)

### LINKED ISSUE #911: Lost modifier / AOBS bug

billbisco:
## Summary
Locks the printed free-fire path for Chewbacca's Bowcaster while it is on Chewie.

## Issue
https://github.com/PlayersCommittee/gemp-swccg-public/issues/999 reported the bowcaster costing 3 Force late in a game (Chewie of Kashyyyk (V)). Related to the lost-modifier class in https://github.com/PlayersCommittee/gemp-swccg-public/issues/911.

## AR / card text
Printed: "May target a character, creature or vehicle using 3 Force (for free if Chewie firing)."

If Chewie is firing, the cost is 0. A 3 Force cost while Chewie is still the firer would mean the free-fire modifier stopped applying.

## Tests
`ChewbaccasBowcasterFiresForFreeOnChewieAtZeroForce` — attached to Chewbacca Of Kashyyyk, 0 Force in the weapons segment, Fire is still offered.

Could not reproduce the late-game 3 Force cost on a short table. This PR does not claim to fix #911; it documents the legal free path so a live 3 Force cost is clearly a lost modifier.

Refs #999 #911.


billbisco:
This PR is superseded by https://github.com/PlayersCommittee/gemp-swccg-public/pull/1102 which already includes the 0-Force Bowcaster tests plus the skip-set engine change.

PR #1102, opened 2026-09-15
Prevent lost modifiers from a skip-set leak during query

billbisco:
## Summary
Engine hygiene for the lost-modifier class (#911), including the Chewbacca's Bowcaster late-game 3 Force case (#999).

This does **not** bake Chewie-free into the Bowcaster fire action. Printed cost stays 3; free fire is still the while-in-play `FiresForFreeModifier`. The change is to stop the engine from dropping modifiers of this kind if a query hiccups.

Replay [TacoBill$c63dcch5ipzik9df](https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=TacoBill$c63dcch5ipzik9df) showed free fire on LS turn 3 and 3 Force on LS turn 6 with the same Chewie and bowcaster still attached. That matches a lost while-in-play modifier, not a rules miss.

**AR** ([2023 Advanced Rulebook](https://res.starwarsccg.org/rules/SWCCG_2023_AdvancedRulebook.pdf), Weapons): pay the listed fire cost; if the weapon fires for free, that cost is 0.

## What this is (and is not)

Not a proof that #999 is reproducible in a normal game. Strong certainty that **if** this class of leak appears, it can be contained.

Every time GEMP asks “does this modifier apply?”, it:

1. Puts the modifier in `_skipSet` (so a condition cannot recurse forever).
2. Evaluates the condition (for Bowcaster: attached to Chewie?).
3. Takes it out of `_skipSet`.

While it is in the set, nested queries pretend that modifier does not exist. That is intentional **during that one query**. After step 3 it should apply again.

Old code had no `try/finally`. If step 2 threw, step 3 never ran. The modifier stayed skipped for the rest of the game. Next Fire used printed 3 Force; right-click Other modifiers omitted the line. Snapshots also copied that in-flight set, so a revert could restore the leak.

The skip-set unit test **forces** that throw on purpose (a synthetic crash, not TacoBill’s sequence). First query blows up; second query must still find `FiresForFreeModifier`. Before the fix, the second query was empty. After, it is not. That is the hole this PR closes.

A six-turn **legal** game (activate, deploy Chewie and Bowcaster from hand, battle each LS turn) still fires for free **with or without** this fix. That sequence never throws inside modifier evaluation, so skip-set has nothing to leak. It is a regression lock that a normal multi-turn game does not drop free-fire by itself. It is not evidence that this change would have saved that replay.

Suggestion: keep #911 and #999 open until a live game confirms the class is gone.

## Test plan
- `ModifiersLogicSkipSetTests.exceptionDuringModifierQueryDoesNotPermanentlySkipModifier`
- `ModifiersLogicSkipSetTests.nestedRemoveOfSameTypeDoesNotPermanentlySkipModifier`
- `ModifiersLogicSkipSetTests.snapshotDoesNotPersistInFlightSkipSet`
- `ChewbaccasBowcasterFiresForFreeOnChewieAtZeroForce`
- `ChewbaccasBowcasterFiresForFreeAcrossSixPlayedTurns`


billbisco:
Clarifying what the skip-set test proves, so this is not read as “we reproduced TacoBill.”

**Not** a proof that #999 is reproducible in a normal game. **Strong certainty** that if this leak appears, it can be contained.

GEMP’s modifier query uses `_skipSet` as a “already looking at this” flag:

1. Put the modifier in the set.
2. Evaluate its condition.
3. Take it out.

Old code had no `try/finally`. If step 2 threw, the modifier stayed skipped for the rest of the game. Bowcaster then used printed 3 Force; right-click Other modifiers omitted “Fires for free.”

The unit test **forces** that throw once, then asks again. Before the fix, the second ask was empty. After, the modifier is still there. Snapshots also no longer copy an in-flight skip-set.

A six-turn legal game (activate / deploy from hand / battle) still fires for free with **or without** this fix. That game never throws mid-query, so it does not exercise the hole. Keep #911 and #999 open until a live game confirms the class is gone.

BOTVHD:
Before trying to fix with this change, is it easy to put in some diagnostic entry that will show that this is what is actually happening?  Like if _skipSet is supposed to be empty at certain times, maybe we can record when that is not the case, confirm that it resulted in the lost modifier symptom, and then start pursuing how/why that is happening?

Also, is this being handled differently in the LOTR gemp where, to our knowledge, the lost modifier thing is not a known issue?

billbisco:
Thank you, VHD.

skipSet must be empty between queries. Added a diagnostic that records (and prints) when it is not, with the leftover modifier type and source. leftoverSkipSetWhenIdleIsRecordedAndModifierIsMissing: a leftover skipSet entry makes the next query miss that modifier (the lost-modifier symptom) and writes the diagnostic.

The try/finally still removes the entry after an exception so a query failure cannot leave it skipped. Snapshots still do not copy an in-flight skipSet.

LOTR GEMP uses the same skipSet recursion guard. The main modifier loop still add/remove without try/finally. hasKeyword does use try/finally. LOTR does not snapshot skipSet, so a revert-copy leak is a SWCCG-only extra hole.

PR #1105, opened 2026-09-16
Include attached cards on Overwhelmed and Ambush hosts

### LINKED ISSUE #967: Overwhelmed / Ambush - Card selection / destination of weapons/devices/effects etc
Yes

## **Describe the bug**
When Overwhelmed / Ambush prompts a player to send cards away, certain behaviors may not align with the rules.
With Overwhelmed, weapons and devices that are attached to a starship are not selectable, even though they should be. This may have to do with the ambiguous definition of "aboard" in rules terms versus what it means in GEMP terms.
Also, Overwhelmed and Ambush are sending Effects (like Special Modifications) to the Used Pile, even though the All Cards situation rules state that Effects are not included in "all cards" and thus might technically be supposed to go to Lost Pile.
Likewise, attached creatures like Mynocks and aboard creatures like Porgs should be looked at.

## **Expected behavior**
Although we know some things are definitely wrong, allow Rules Team to discuss and come up with concrete specifications before any changes are made.

## **Additional context**
Low priority. Related to #524 but decided to break this out into a separate issue.

billbisco:
Why: leftover #967. Overwhelmed did not let you select weapons/devices attached to a starship, and Overwhelmed/Ambush were sending attached Effects to Used Pile.

What it Does: Overwhelmed and Ambush collect each host plus cards on them (weapons, devices, Effects, creatures), not only pilots/passengers/cargo. Placement is Used Pile. Adds tests: attached Rectenna and Special Modifications on Overwhelmed go Used and are selectable; Ambush attached Comlink goes Used, not Lost.

What it Fixes: Attached devices/Effects on the host are selectable and go Used. Printed "cards on them" is Used Pile, not the All Cards Lost-Pile default.

Issues: Closes #967


billbisco:
Printed Overwhelmed: "Place all opponent's starships there (and cards on them) in owner's Used Pile." Ambush is the same pattern for characters, vehicles, and starships at a site.

AR All Cards/All Characters applies to the phrase "all cards." These interrupts do not say "all cards." They name hosts and add "cards on them," with Used Pile as the destination.

Weapons/devices attached to the host should be selectable and go Used. Effects such as Special Modifications are "cards on them" and go Used. Attached creatures go with the host. Overwhelmed is still illegal if the opponent has a Jedi or starship weapon at the system; Ambush is still illegal if the opponent has a Dark Jedi or character weapon at the site.

PR #1111, opened 2026-09-16
See-Threepio retrieves Force before the replaced persona is Lost

### LINKED ISSUE #894: See-Threepio retrieval happening after persona replacement

## **Describe the bug**
Currently, See-Threepio's force retrieval occurs immediately after the persona replacement (after the other persona has been sent to lost pile).  It should be happening before persona replacement:

From discussion with Gergall:
"When replacing another C-3PO" should be read as "When about to replace another C-3PO" (not "If he just replaced another C-3PO")

The closest AR citation:
_For complex timing interactions, consider the phrase "leaving the table" to be functionally equivalent to "about to leave the table."_

## **To Reproduce**
1. Deploy a non-See-Threepio persona of C3P0
2. Persona replace with See-Threepio
3. Observe that the old C3P0 persona card is incorrectly retrieved

## **Additional context**
Rework needs to change getGameTextRequiredBeforeTriggers into getGameTextRequiredAfterTriggers and use a (new) isReplacingPersonaCharacter trigger.

Currently fails test SeeThreepioRetrieves3Force() in src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set110/light/Card_110_003_Tests.java

billbisco:
Why: leftover #894. Printed See-Threepio retrieves 3 Force when replacing another C-3PO. That is during replacement, not after the old persona is already in Lost Pile.

What it Does: Adds an about-to-replace result while the replaced card is still not Lost. See-Threepio fires on that. The existing just-replaced result stays after Lost Pile. Tests retrieve 3 without retrieving the replaced C-3PO.

What it Fixes: See-Threepio no longer retrieves the C-3PO he just replaced. A later just-replaced trigger still sees the old persona in Lost Pile.

Issues: Closes #894


BOTVHD:
This would work, but I don't think it is a good solution.

Announcing that the persona replacement was just completed (the result emitted) should stay after the last step (putting the replaced card in the lost pile).

Imagine a future card that said "if you persona replaced a character, put the replaced character in hand" - or out of play, or whatever.  In the proposed change, that would break things because after acting on the response, the replaced character would be sent to the lost pile (as the final step in character replacement).

billbisco:
Thank you, VHD. You are right that "just persona replaced" should stay after Lost Pile.

See-Threepio now fires on an about-to-replace result while the old C-3PO is still not in Lost Pile. The existing just-replaced result is after Lost Pile again.

PR #1112, opened 2026-09-16
A Fine Addition lets Grievous use a persona-restricted saber

### LINKED ISSUE #889: A Fine Addition To My Collection not letting Grievous swing the lightsaber
Yes. This was also reported in #833 but I'm splitting this off so that we don't have two unrelated bugs in the same bug report.

## **Describe the bug**
A Fine Addition To My Collection will deploy a lightsaber on Grievous, and it will give him a visible modifier that says he can use it... but then he still cannot use it.

## **To Reproduce**
1. Lose Vader's Lightsaber
2. Play A Fine Addition to deploy Vader's Lightsaber from Lost Pile onto Grievous
3. Initiate battle at Grievous' location
4. Grievous has an appropriate modifier, but cannot swing the saber.

## **Game Replay ID:**
Here is the replay from the #833 report:
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=Bottman$7u6kx36bky9iqx0f

## **Expected behavior**
Grievous should be able to swing the saber.

## **Screenshots**
This screenshot tells the full story. Chat log shows battle just initiated. Grievous has a modifier from A Fine Addition. The saber is not lighting up to swing.
Image

billbisco:
Why: leftover #889. Printed A Fine Addition: deploy any lightsaber from Lost Pile on Grievous (he may use it until he is no longer carrying it).

What it Does: Records the grant with WhileInPlayData on the deployed saber. MayUseWeaponModifier is gated by that data, the saber's permanent card id, and attachment to Grievous. A table-changed required action clears the data the first time the saber is no longer attached to him. Tests: stats, deploy from Lost Pile, Grievous can use a Dark Jedi saber and Vader's Lightsaber, and transferring the saber away does not restore use.

What it Fixes: Grievous can use a persona-restricted saber deployed by A Fine Addition. The grant does not return if the saber leaves and comes back.

Issues: Closes #889


BOTVHD:
This is probably the correct fix for the described bug.

I think there is a partially related bug, though - I don't see anything that enforces the gametext: "(he may use it **until he is no longer carrying it**)"

As written, it looks like Grievous could no longer be carrying the saber (returned to hand, stolen, sent to lost pile) and then (incorrectly) resume using it (deploy it, steal it back, etc).

We need a way to remove the MayUseWeaponModifier when the target saber is no longer attached to Grievous.

billbisco:
Thank you, VHD. The grant now ends the first time the saber is no longer attached to Grievous, so stealing it back does not restore use.

BOTVHD:
It may be worth re-working the "until no longer carrying it" to use the WhileInPlayData features.  When the action is taken, record the lightsaber card by appending the effect:
`new SetWhileInPlayDataEffect(action, self, new WhileInPlayData(lightsaber))`
and use the InPlayDataEqualsCondition to gate the MayUseWeaponModifier.

Then have an automatic (required) action that sets the text string to null if the getWhileInPlayData cardId is no longer attached to self (maybe using a getGameTextRequiredAfterTriggersAlwaysWhenInPlay with a condition of TriggerConditions.isTableChanged()?).

I think that should be more robust (and may handle reverts better, since the snapshot data will track the InPlayData?).

See example Card13_045 that saves a lightsaber this way.

billbisco:
Thank you, VHD. WhileInPlayData now records the grant. A Fine Addition is a Lost Interrupt, so that data lives on the deployed saber rather than on the interrupt. MayUseWeaponModifier is gated by InPlayDataSetCondition and the saber's permanent card id while attached to Grievous. A table-changed required action clears it the first time the saber is no longer attached to him.

PR #1114, opened 2026-09-16
Move-as-react embark happens after the cancel window

### LINKED ISSUE #872: Lift Tube Escape embarking before response to cancel 'react'

## **Describe the bug**
https://forum.starwarsccg.org/viewtopic.php?p=1435024#p1435024

Lift Tube Escape is played, characters embark on the Lift Tube, move, then a react or interrupt canceler, such as Sense, is played, forcing the Lift Tube back to the original site but trapping all characters inside it, leading to a lopsided battle.

## **Game Replay ID:**
Unavailable? From 2025 OCS game with bsilkwoo and jayjaydee

## **Expected behavior**
Lift Tube Escape is played, and at that point it can be canceled, before any embarking or movement is actually done by the Lift Tube.

Workaround: revert to a game state before the playing of Lift Tube Escape, move the Lift Tube using the interrupt without passengers aboard, and then allowing the cancellation to occur.

## **Additional context**
Likely not limited to Lift Tube Escape / Lift Tube - probably an issue with 'react' embarking and response timing?

billbisco:
Why: leftover #872. Move-as-react was embarking passengers before Sense could cancel the react.

What it Does: Opens the cancel window after the move cost and before embark. Inner land/takeoff/move types stay unset so Sense is not offered a second time. Tests that cancel leaves passengers on the ground.

What it Fixes: Canceling the react leaves the reacting card and would-be passengers where they were.

Issues: Closes #872


BOTVHD:
I'm concerned about removing all of the getType() calls returning the correct type (ex: Type.TAKING_OFF_AS_REACT ) under react conditions...  Why are all of those being removed and won't that interfere with the ability of MoveAsReactAction displaying the correct text, triggerConditions.isReact() returning the correct boolean, etc?

billbisco:
Thank you, VHD. I looked again.

Inner land/takeoff/landspeed effects keep getType null on purpose so Sense is not offered a second time during the actual move. InitiateMoveAsReactEffect still returns TAKING_OFF_AS_REACT (and the related types) for the cancel window. Display text still uses the as-react flag.

Restoring those inner types fails ComScan Detection and Flagship move-as-react tests.

PR #1115, opened 2026-09-16
Lost Pile order from table is chosen by the player whose turn it is

### LINKED ISSUE #858: Explosive Charge has LS choosing the order of lost characters.

## **Describe the bug**

When Explosive Charge explodes, causing all cards to be lost, LS chooses the order of the lost cards. It should be DS, because this trigger can only occur during DS’s turn, meaning DS gets to choose the order of competing automatic actions. 

## **To Reproduce**

Have LS put an Explosive Charge at an interior site, then have DS initiate and win a battle on a subsequent turn. 

## **Game Replay ID:**

https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=bibfortuna$6tl433vbyxe7xosd

Dark side turn 4. 

## **Expected behavior**

Dark side should choose which characters go lost in which order.

billbisco:
## Why

[#858](https://github.com/PlayersCommittee/gemp-swccg-public/issues/858) reports that when Explosive Charge explodes, Light Side chooses the order cards go to the Lost Pile even though it is Dark Side's turn.

AR p.11: when an action places multiple cards into a hand, deck, or pile and the card text does not name a player, placing each card is a competing automatic action, and the player whose turn it is chooses the order.

`LoseCardsFromTableEffect` asked `action.getPerformingPlayer()` (the Charge's owner) instead of the current player. Same hole would show on Unfriendly Fire, Thermal Detonator, etc. when the owner is not the player whose turn it is.

## What it does

- `ChooseNextCardToLose` uses the current player.
- Test: LS-sourced multi-card lose during DS turn; DS is asked "Choose card to be lost".

## What it fixes

Lost Pile order for Explosive Charge (and other table-loss piles using this effect) is chosen by the player whose turn it is.

## Issues

Closes #858


BOTVHD:
This also changes the way Monnok (lost) handles card loss, right?  When played during LS turn, allows LS to choose the order.  When played during DS turn, allows DS to choose the order?

Referencing the quoted section of the AR... for cards that do "name a player", does this still work correctly?

There may be some additional effects that need the same change (choice goes to active turn player instead of action performer).

billbisco:
Thank you, VHD.

Monnok (Lost) is not this PR. It uses LoseCardsFromHandEffect, which still asks the performing player. This change is only LoseCardsFromTableEffect.ChooseNextCardToLose.

Explosive Charge (and other table multi-card loses on this path) does not name who chooses Lost Pile order, so AR p.11 gives that choice to the player whose turn it is. A card that did name a player for that order would need its own path; this effect does not take a named-player override.

ChooseCardsToLoseFromTableEffect.ChooseNextCardToLose, LoseCardsFromHandEffect, and LoseCardsFromSabaccHandEffect still use the performing player. Same AR would apply there when the text does not name a player.

PR #1117, opened 2026-09-16
Honor Transmission Terminated (V) when using tractor beams

### LINKED ISSUE #846: Transmission Terminated(v) targeting a ship to prevent tractor beams did not protect from In Range
DS player deployed Tyrant and LS player played transmission terminated(v) targeting tyrant to not be able to use tractor beams.

DS player decided to YOLO and battled anyway and In Range highlighted. Testing for bugs In Range was played in the weapons segment and successfully fired the tractor beam.

https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=Teacher$7jp7gk9qpjsuu1g6

## **Expected behavior**
Tyrant should not be able use the tractor beam if targeted by transmission terminated(v).

billbisco:
Why: leftover #846. Transmission Terminated (V) targets a starship so that, for the rest of the turn, it may not use tractor beams, fire weapons, or cloak. In Range still offered and resolved a tractor beam on a targeted Star Destroyer.

What it Does: Adds Filters.weaponOrDeviceMayBeUsed so tractor-beam actions skip devices that may not be used. In Range, Captain Lennox, and Tractor Beam's end-of-battle trigger check that restriction. UseTractorBeamEffect does not resolve if the device may not be used. Tests: regular Tractor Beam, In Range, and Lennox are all prevented after TT(V).

What it Fixes: After Transmission Terminated (V) targets a Star Destroyer, that ship cannot use its tractor beam for the rest of the turn, including In Range and Lennox.

Issues: Closes #846


src/gemp-swccg-server/src/test/java/com/gempukku/swccgo/cards/set223/light/Card_223_006_Tests.java
```diff
@@ -86,60 +86,40 @@ public void TransmissionTerminatedVStatsAndKeywordsAreCorrect() {
...
-    @Test @Ignore
-    public void TransmissionTerminatedVUsedPreventsUsingTractorBeam() {
-        //test coverage:
-        //test1: can target opponent starship
-        //test2: can target self starship
-        //test3: (FAILS) opponent targeted starship cannot use tractor beam this turn
-        //test4: opponent targeted starship can use tractor beam next turn
-
+    @Test
+    public void TransmissionTerminatedVUsedPreventsInRangeUsingTractorBeam() {
```
BOTVHD, line 90: I think this should be a separate (additional) test instead of replacing the one that currently exists.  We do want to show that TT(V) is working correctly for just a regular Tractor Beam firing.  

The new test with In Range is equivalent to first suggested additional test in comments at the bottom of this test file.

Adding a third test that confirms Lennox is prevented wouldn't hurt, either (the second suggested additional test in comments at the bottom of this test file).

BOTVHD:
At a glance, this seems reasonable.  It's interesting that this new weaponOrDeviceMayBeUsed check on 'self' hasn't been necessary for other weapons/devices.  That makes me wonder if the 'may be used' checking feature is supposed to be handled at a higher level.

Unrelated, but I suspect there are a number of other isPlayableInFull overrides for various effects that return 'true' instead of checking correct conditions - but that is outside this scope.

src/gemp-swccg-cards/src/main/java/com/gempukku/swccgo/cards/set2/dark/Card2_115.java
```diff
@@ -76,6 +76,7 @@ private Filter getTargetFilter(PhysicalCard self) {
     protected List<OptionalGameTextTriggerAction> getGameTextOptionalAfterTriggers(String playerId, SwccgGame game, EffectResult effectResult, PhysicalCard self, int gameTextSourceCardId) {
         // Check condition(s)
         if (TriggerConditions.battleEndingAt(game, effectResult, Filters.here(self))
+                && Filters.weaponOrDeviceMayBeUsed.accepts(game, self)
```
BOTVHD, line 79:
```suggestion
                && GameConditions.canUseDevice(game, self)
```
This matches the way most of the rest of the devices cards gate their actions

billbisco:
Thank you, VHD. The regular Tractor Beam case is back as its own test. In Range stays as a second test, and there is now a third for Captain Lennox. Tractor Beam's end-of-battle trigger still uses weaponOrDeviceMayBeUsed; that is the filter that hides the device after Transmission Terminated (V).

PR #1118, opened 2026-09-16
Blown-away locations cause their own Force loss

### LINKED ISSUE #837: Honor Of The Jedi not reducing Force loss from DS Hoth Main Power Generators
Yes

## **Describe the bug**
DS blows away the DS card Hoth: Main Power Generators from Set 22 (Card222_009.java).  DS does not satisfy Honor, occupying only a single battleground. However, the full Force loss of 5 is applied. Of note: No Escape is on table, but No Escape should not be applicable because the Force loss comes from a location card, not from the Epic Event.

## **To Reproduce**
1. Have DS play the Hoth Walkers objective
2. LS deploys Honor Of The Jedi
3. Possibly relevant: Have DS deploy No Escape
4. DS blow away Main Power Generators without satisfying Honor
5. The full 5 Force loss occurs

## **Game Replay ID:**
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=wcullen$5qjuh2yjtqloh8oq
DS Turn 4

## **Expected behavior**
Honor should reduce the loss from 5 to 3.

## **Screenshots**
Image

billbisco:
## Why

Honor Of The Jedi did not reduce the 5 Force loss when Dark Side Hoth: Main Power Generators (Set 22) was blown away. The location text is “If 'blown away,' Light Side loses 5 Force,” which is a location-sourced Force loss, not “+5” added to the epic event’s blown-away total. Folding it into the epic event made No Escape treat the loss as coming from an Epic Event, which is immune to Honor.

## What it does

- Main Power Generators (Set 22 and original) cause their own `LoseForceEffect` sourced from the location.
- Same pattern for Death Star: War Room (V) (“you lose 3 Force”) and Blockade Flagship: Docking Bay (“lose 4 Force”), which also used the additive blown-away pool without a “+”.

## What it fixes

Blowing away Main Power Generators is a 5 Force loss from the location. Honor reduces it by 2 unless the opponent occupies three battlegrounds. No Escape does not apply, because the source is a location, not an Effect, Epic Event, or Objective.

## Issues

Closes #837


BOTVHD:
Because this force loss is not directly part of the blow away action, I'm not totally sure if the site's force loss should occur as a response to the TriggerCondition.isBlownAwayCalculateForceLossStep.  It's possible it should be happening at a different point in the sequence (like, for example, the TriggerCondition.isBlownAwayLastStep).

Does the AR clearly specify the timing for force loss from location text (that is not modifying the blown away amount of force loss)?

billbisco:
Yes. 2023 AR Appendix C, **Blown Away - Timing**: after the card says the location is blown away, in order:

1. Relocate cards that relocate when blown away.
2. If a site, cards at that site are lost.
3. **Any Force loss caused by any cards or rules is now checked and applied.**
4. If a system, related locations (and cards there) are lost.
5. Flip the location.
6. Continue with events on the card.
7. Automatic actions to the blowing away.
8. Just actions to the blowing away.

Step 3 is all Force loss from cards or rules, not only a "+" to the epic-event total. Example 2 (Bunker): at step 3 "the Force loss on the Epic Event is now applied ... we check all sources." Example 1 (Hoth system): the Main Power Generators 8 is not applied because that *site* was not blown away; if it had been, that 8 is Force loss at this same step.

That is `isBlownAwayCalculateForceLossStep` (after cards at the site are lost, before the flip). `isBlownAwayLastStep` is AR steps 7-8, after the location is already face-down with no printed game text. Location text "If 'blown away,' Light Side loses 5 Force" is checked at step 3, as its own LoseForceEffect from the location.

BOTVHD:
I think this is right.

I'm not totally sure about the removal of `action.skipInitialMessageAndAnimation();`
I guess that's a judgement call on UI (forcing the player to take a required action on the location to do the force loss action instead of auto-prompting it?).

I'm not very familiar with how the performing player affects things, but the addition of `action.setPerformingPlayer` seems reasonable.

billbisco:
`skipInitialMessageAndAnimation` only skips the initiation message and card animation. Required triggers still auto-fire either way.

It belonged with `AddToBlownAwayForceLossEffect`, which was a silent add to the epic-event pool. This is now a real `LoseForceEffect` from the location, so that message should show.

`setPerformingPlayer` is for that message (who is applying the location text). Thanks VHD.

PR #1131, opened 2026-09-16
Offer Fenn Rau retrieve only after activating 2 Force

### LINKED ISSUE #766: Fenn Rau retrieving when not allowed

## **Describe the bug**
Fenn Rau prompts LS for retrieval even when 2 Force was not activated when returning to hand.

## **To Reproduce**
1. Return Fenn Rau to hand with 1 card in Reserve Deck.
2. Activate 1 Force
3. Retrieve 1 Force

## **Game Replay ID:**
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=jokerking$tp73ksr3mkrxxqyj

Light Side Turn 5

## **Expected behavior**
LS activates 1 Force, but cannot retrieve due to not completing the 2 Force activation (confirmed with Gergall)
.

billbisco:
Why: leftover #766. Printed Fenn Rau: return to hand to activate 2 Force (if he won a battle this turn, may also retrieve 1 Force). Retrieve is additional to completing that activate 2.

What it Does: The optional action is not offered unless Reserve has at least 2 cards. Retrieve waits until activate 2 actually happened. Tests that 1 card in Reserve does not offer the action, and that retrieve is offered after activating 2 when he won the battle.

What it Fixes: Fenn Rau is not offered when he cannot activate 2 Force, and he no longer retrieves unless that activate completed.

Issues: Closes #766


BOTVHD:
Doesn't this introduce a new bug by allowing the player to choose the optional action when there are 0 or 1 cards in Reserve Deck?  Why was the requirement of having 2 cards in reserve deck (as a pre-req for presenting the action) removed?

billbisco:
Thank you, VHD. The optional action is not offered unless there are at least 2 cards in Reserve. Retrieve still waits until activate 2 actually happened.

BOTVHD:
> Thank you, VHD. The optional action is not offered unless there are at least 2 cards in Reserve. Retrieve still waits until activate 2 actually happened.

I think I was unclear.  I was talking about the retrieve action requirement.  The GameConditions.canActivateForce is probably the correct way to evaluate if the the player can take the activation action (example: Put That Down (V) ).  I didn't see the check against 2 cards after the change (gating retrieval), but I think that is superseded by the new wasCarriedOut() check.

As another example, Figrin can attempt the retrieval action as long as both players have 1 force in Reserve (even though some of the draws will probably fail when attempting to draw multiple destiny).

billbisco:
Thank you, VHD.

Retrieve is not gated on 2 remaining in Reserve. After return-to-hand, retrieve is offered only if activate 2 was carried out (and he won a battle this turn). Same idea as Figrin: the retrieve is additional after that activate completed, not a second 2-card Reserve check.

The action itself still requires canActivateForce and 2 cards in Reserve so return-to-hand to activate 2 Force is not offered when that activate cannot happen. That is the playability check; wasCarriedOut is the retrieve check.

Issue #760, opened 2024-04-15
Dagobah Bog Clearing allows Captain Han Solo to deploy on the falcon as a single action to the bog clearing


charlesmhickeyjr:
- [ ] Did you search through the list of **open issues** _first_ to ensure that there is not already an report for this issue?
No

## **Describe the bug**
Millenium Falcon and Captain Han Solo can deploy to the dagobah bog clearing as a single action.

## **To Reproduce**
Deploy Falcon with Captain Han to the bog clearing.

## **Game Replay ID:**
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=Teacher$1tcg9gpni899i2iu

## **Expected behavior**
Han should no be able to deploy to dagobah,

### **Desktop (please complete the following information):**
 - OS: [e.g. iOS] ios
 - Browser [e.g. chrome, safari] chrome
 - Version [e.g. 22]  122.0.6261.112

Gergall:
As this has come up a few times, I just wanted to state that the rules team confirms this is a bug. Captain Han should not be able to do that.

billbisco:
Printed Captain Han Solo: deploys only on Falcon, Hoth or Cloud City. Printed Bog Clearing: your starfighters may deploy here.

Falcon can land at Bog Clearing. Han still cannot deploy to Dagobah. Simultaneous Falcon + Han to Bog Clearing is Han deploying to Dagobah, which the rules team confirmed is illegal.

Simultaneous starship/pilot deploy now still applies the Dagobah (and Ahch-To) prohibition to the pilot even when that pilot may board the ship at any location.

PR #1136, opened 2026-09-16
Stop Trash Compactor from unlocking Death Star system for ship plus pilot

### LINKED ISSUE #736: Trash Compactor affects Death Star system

billbisco:
## Why

Death Star: Trash Compactor was reported as letting Light Side simultaneously deploy an empty starship and a pilot to the Death Star system. Permanent-pilot X-wings stayed blocked, so the hole was the ship-plus-pilot path.

Printed Trash Compactor Light text: you may deploy here without presence. If you control, Force drain +1 here.

"Here" is the Trash Compactor site. The related Death Star system is a different location and still requires presence or Force icons.

## What it does

Stops promoting a location-scoped "deploy without presence" modifier into a blanket DeploymentRestrictionsOption during simultaneous starship-plus-pilot deploy. Presence/icon exceptions stay per-target.

Adds tests that Han may deploy to Trash Compactor but not the system, that an X-wing cannot deploy to the system, and that Falcon + Captain Han cannot simultaneously deploy to the Death Star system but still can to a Cloud City docking bay.

## What it fixes

Empty Falcon + pilot can no longer deploy to the Death Star system just because Trash Compactor is on table.

## Issues

Closes #736


BOTVHD:
Is this supposed to be instead of https://github.com/PlayersCommittee/gemp-swccg-public/pull/1069 ?  It seems like both are talking about fixing the same problem?

BOTVHD:
I don't see how this change should fix the issue.

It avoids adding some conditional relaxed deployment options if the card has a MAY_DEPLOY_WITHOUT_PRESENCE_OR_FORCE_ICONS modifier.  But that modifier is only applied by  MayDeployToTargetWithoutPresenceOrForceIconsModifier - and the only card in the game that applies that is Tydirium.  So changes to that piece of code (I would think) should only affect Tydirium-related deployment.

Are these new tests failing before applying the fix?

billbisco:
Thank you, VHD.

Same leftover #736 as #1069. #1069 is the PR for that issue (and also #760). This PR is the Trash Compactor half of that same AbstractStarship change, not a replacement for #1069.

The modifier type is not Tydirium-only. Trash Compactor uses MayDeployToLocationWithoutPresenceOrForceIconsModifier, which is the same MAY_DEPLOY_WITHOUT_PRESENCE_OR_FORCE_ICONS type and affects your cards not in play. The old simultaneous ship-plus-pilot loop promoted any such modifier into a blanket DeploymentRestrictionsOption, so Falcon + Captain Han was offered the Death Star system.

Yes: TrashCompactorDoesNotAllowFalconAndPilotSimultaneousDeployToDeathStarSystem fails on the unfixed AbstractStarship (Death Star is offered). After the change it is not.

PR #1142, opened 2026-09-16
Offer Mandalorian Mishap (V) destiny cancel once after drawing completes

### LINKED ISSUE #353: Mandalorian Mishap (v) triggering twice
Expected behavior: Opponent plays Mandalorian Mishap (v) to cancel 1 BD
Actual behavior: Interrupt triggered twice, which forced LS to use 2 force since First Strike was on table.

I have asked opponent to post his game replay link so you are better able to see the precise issue from his end.  My game replay is: https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=jokerking$v673rnm8hlh05xfg

billbisco:
## Why

Mandalorian Mishap (V) was played twice in one battle (nested), paying Force twice, after opponent drew three battle destinies. Printed text: during a battle, if opponent just drew more than two battle destiny, cancel one of those destiny draws.

That is one response after drawing is complete, not a repeatable response for the rest of the battle.

## What it does

The optional now requires battle-destiny drawing just completed for the opponent, matching the same gate used by similar interrupts.

## What it fixes

MM (V) no longer initiates playing twice for the same three-destiny draw.

## Issues

Closes #353


BOTVHD:
This change makes matches the way didDrawMoreThanBattleDestinies is used in the two other cards:
- Card10_028 (Abyssin Ornament & Wounded Wookiee)
- Card6_161 (Wounded Wookiee)

However, these other cards use the phrase "if opponent drew more than" instead of Mishap (V)'s phrase "if opponent **just** drew more than".

I'm not sure if that warrants a different timing (allowing a response before all destinies have been drawn?) or if all 3 of these cards are functionally intended to work the same way.

Probably should clarify with rules team how this one is supposed to work.

billbisco:
Thank you, VHD. Printed Mandalorian Mishap (V) is "if opponent just drew more than two battle destiny." "More than two" is only known after that drawing is complete, and "just" is that completed-drawing window, not a later time in the battle and not a response after the second draw while more destinies are still being drawn.

Wounded Wookiee / Abyssin Ornament omit "just" but use the same complete-draw count. The "just" here still maps to isBattleDestinyDrawingJustCompletedForPlayer plus didDrawMoreThanBattleDestinies, same complete-draw check those cards use.

PR #1067, opened 2026-09-12
Fix #201: stop battle destiny after Program Trap clears presence

billbisco:
## Summary
Stops further battle destiny draws when a battle loses presence mid-power-segment (e.g. Program Trap exploding after a destiny draw). Closes #201.

## What changed
`BattlePowerSegmentAction` previously queued the whole power-segment destiny sequence up front. Presence can disappear mid-sequence (Program Trap, similar clears), but later battle destiny draws still ran.

Each queued power-segment destiny step is now wrapped in a `canContinue` gate (same idea as `BattleCheckIfWeaponsSegmentFinishedEffect` in the weapons segment). If the battle can no longer continue, remaining destiny draws / attrition / initial attrition calculation are skipped.

## Tests
`Card_2_124_Tests` — **2 run, 0 fail, 0 skip**
- `ProgramTrapStatsAndKeywordsAreCorrect`
- `ProgramTrapExplodingMidPowerSegmentStopsFurtherBattleDestinyDraws`

## Out of scope
Does not rewrite `BattleEffect` segment orchestration or address cancelled-battle immediacy (#691).

PR #1144, opened 2026-09-17
Stop remaining battle destiny draws after Program Trap clears the site

### LINKED ISSUE #201: BUG: Battle destiny drawn even after Program Trap explodes at battle site during battle.
Screen Shot 2020-11-06 at 12 39 35 PM

Game link: https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=CRG$atnnz5ta90fz1fx4

Game log: 

shadowb draws •Meson Martinet as a 3 for battle destiny
•Program Trap required response is initiated
•Program Trap targets to make droid 'explode'
•Program Trap 'explodes'
shadowb causes •Luke Skywalker, The Rebellion's Hope, •Luke's Blaster Pistol (V) on table to be lost using •Program Trap
CRG puts •Luke Skywalker, The Rebellion's Hope on Lost Pile
CRG puts •Luke's Blaster Pistol (V) on Lost Pile
shadowb causes •Chewie With Bowcaster on table to be lost using •Program Trap
CRG puts •Chewie With Bowcaster on Lost Pile
shadowb causes •IG-88 With Riot Gun, •Program Trap on table to be lost using •Program Trap
shadowb puts •IG-88 With Riot Gun on Lost Pile
shadowb puts •Program Trap on Lost Pile
shadowb's total battle destiny is 3
CRG draws •Lando With Vibro-Ax as a 1 for battle destiny
CRG's total battle destiny is 1
Battle at •Tatooine: Jabba's Palace ends

DS turn 5, battle phase. 

**WHAT:** Program Trap is deployed on EPP IG-88. DS initiates battle at IG'88's site. Program Trap (properly) explodes after a destiny draw, sending all characters in battle lost. Despite this, LS then drew battle destiny. (No battle winner was determined, and so no battle damage was calculated).

**EXPECTED:** Battle should automatically end when all characters who were participating in battle are removed (just as if one side no longer had any cards with ability or which gave presence). LS should not then draw battle destiny.

billbisco:
## Why

Program Trap exploding after the first battle destiny still let the opponent draw battle destiny even though every character at the site was already lost.

Printed Program Trap: when either player draws a destiny matching the number of characters at same site, the droid explodes and all characters present are lost.

A battle that no longer has presence for both sides during the power segment is over. Remaining destiny draws must not happen.

## What it does

Each power-segment destiny draw (power-only, battle destiny, attrition-only) and the initial attrition calculation now skip if the battle can no longer continue.

## What it fixes

After Program Trap clears the site mid-power-segment, Light Side is not asked to draw battle destiny and the battle does not continue.

## Issues

Closes #201


BOTVHD:
Duplicate?

billbisco:
Duplicate of #1067. This one does not supersede it.

#1067 already Closes #201 and is the PR for that issue. It gates remaining power-segment destiny draws / attrition on `canContinue` in `BattlePowerSegmentAction`, with `Card_2_124_Tests`.

This PR is the same engine change without tests. Closing it so #201 has one PR.

PR #1071, opened 2026-09-12
Fix Effect relocate deploy-target and immunity checks (#241)

billbisco:
## Summary
Surprise (and Effect relocation generally) could relocate an Effect onto a location that was not a legal deploy target for that Effect ? e.g. Presence Of The Force onto a Yavin 4 site while Yavin 4 (V) blocked DS Effects on related locations. Rules confirmed this is incorrect (Appendix B ?I Have A Bad Feeling About This? style: the relocate primary target must be a legal deploy target).

Also locks immunity: a location immune to an Effect title is an illegal deploy target and therefore an illegal Surprise/relocate target (Expand The Empire ? Scarif: Turbolift Complex; Revolution ? Naboo: Boss Nass' Chambers).

## What changed
- `MayNotDeployToTargetModifier` now affects in-play cards as well (previously `not(in_play)` meant relocate of an already-attached Effect ignored deploy-to-target restrictions).
- `AbstractEffect.getValidRelocateEffectTargetFilter` adds `notProhibitedFromDeployingTo` alongside existing `canBeTargetedBy` (immunity).
- `Filters.canRelocateEffectTo(PhysicalCard)` leaf explicitly checks `isProhibitedFromDeployingTo` and `canBeTargetedBy` (Filter overloads still call this leaf ? no circular recursion).

## Design
- Shared Effect-relocate eligibility path used by Surprise; no per-interrupt rewrites.
- Circular-filter care: only the `PhysicalCard` leaf was extended; `canRelocateEffectTo(Filter)` / `effectCanBeRelocatedTo` remain thin wrappers over that leaf.
- Immunity: encode via `canBeTargetedBy` (ImmuneToTitle) on relocate, same as deploy-to-attached.

## Tests
Class: `Card_5_156_Tests` (3 tests)
- `SurpriseCannotRelocatePresenceOfTheForceToYavinSiteBlockedByMayNotDeployToTarget` ? illegal under Yavin 4 (V); legal relocate to marketplace succeeds
- `SurpriseCannotRelocateRevolutionToBossNassChambersImmuneToRevolution` ? Decipher immunity
- `SurpriseCannotRelocateExpandTheEmpireToScarifTurboliftComplexImmuneToExpandTheEmpire` ? Expand The Empire immunity

Closes #241

Independent of other open PRs (skips #46/#38 cumulative work).

## Known gaps
[...]

PR #1145, opened 2026-09-17
Block Surprise from relocating Effects onto locations they could not deploy to

### LINKED ISSUE #241: Bug - Surprise and Relocating Conditions
[https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=fungineer$j5iegjc2i4dvq24q](url)

Bug - DS Turn #7 
Actual Behavior - DS player was able to use surpise to relocate Precense of the force on Mustafar: Vader's Castle to Yavin IV: War Room. 

Expected behavior. Due to Yavin IV (V) LS being on table, (with DS text stating "unless you control you may not deploy effects on related locations), is would be expected that the relocated effect could not be sent to any Yavin IV sites, as it could not otherwise deploy on them. 

Rules team discussed this (thus is the delay in reporting) and confirmed this should be a bug. See [https://forum.starwarsccg.org/viewtopic.php?p=1180105#p1180105](url) We feel the appendix b entry on "I have a bad feeling about this" appropriately covers the requirements that the effect would have to be able to be able to deploy on the location to be a valid primary target for relocation. 

Though this example bug is specific to surprise and Yavin IV (V) system, it is likely something that should be checked with both relocating interrupts for all existing effects/interrupts.

billbisco:
## Why

Surprise relocated Presence Of The Force onto Yavin 4: War Room while Yavin 4 (V) was on table and Dark Side did not control the system.

Printed Yavin 4 (V) Dark text: unless you control, you may not deploy Effects on related locations.

Relocating an Effect onto a location is still putting it there. That restriction is not only for cards coming from hand.

## What it does

May-not-deploy-to-target now applies to cards already in play, and Effect relocate uses the same deploy-to-target check as deploy.

Adds tests that Surprise cannot send Presence Of The Force to Yavin 4: War Room under that restriction, while a card that is allowed to deploy there still can.

## What it fixes

Surprise cannot park an Effect on a Yavin 4 site the Dark player could not deploy it to.

## Issues

Closes #241


BOTVHD:
Is this a re-do of https://github.com/PlayersCommittee/gemp-swccg-public/pull/1071 ?  As mentioned in another PR, it seems like you are starting to submit multiple PR fixes for the same bug?

billbisco:
Closing as a duplicate of #1071.

#1071 already Closes #241 (opened 2026-09-12). This PR was a second copy of the same Surprise / Effect-relocate deploy-target fix. One PR per issue; #1071 is the PR for #241.

PR #1146, opened 2026-09-17
Restore the converted location when the converting site leaves the table

### LINKED ISSUE #176: Objectives destroy converted sites when removed due to requirements missing
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=umuri$p9k4hmrujh3uhdxa
https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=umuri$e27tia77sw5jcibz

Bug is triggered when dark side deploys a location, and a light side objective deploys a copy of the same location, but cannot complete due to missing some other requirement (in the real game, he was missing a rep, in the test game, missing a generic site).

Expected behvaior when this occurs:
Objective pulls back played cards, which brings the converted location to the top.

Actual behavior:
Objective pulls back played cards, and does not show the converted location at all.

Further analysis:
In the real game, you see a mobile effect (Sandwhirl) travels to the missing site (audience chamber) later on in the game, even though we can't see it, and it messes up the visibility of sandwhirl. We think this is what caused the game to be unplayable later.
So odds are the converted site is still "there", just the topmost entry for it is empty rather than bringing the converted site up in the stack.

billbisco:
## Why

When a Light Objective converted a Dark location and then failed a mandatory start-of-game requirement, undo put the converting site back in Reserve and left an empty location stack. The original converted site was still there but not visible. Later effects (Sandwhirl) still treated it as present.

AR: an illegal Objective is placed out of play and you undo what it did. The converted location must come back to the top of that stack.

## What it does

When a top location leaves the table, the converted location underneath is promoted to the visible Locations zone (it was incorrectly only shown if it was already in that zone).

Adds tests: Agents In The Court with no Rep converts DS Audience Chamber, fails, and DS Audience Chamber is the visible top location again.

## What it fixes

Failed Objective undo no longer hides the converted site.

## Issues

Closes #176


BOTVHD:
Is this supposed to replace, improve, or be in addition to this PR: https://github.com/PlayersCommittee/gemp-swccg-public/pull/1086 (that already claims to fix and close the associated issue)?

It seems like we are starting to get multiple PR submissions for the same fix from AI under your fork?

billbisco:
Duplicate of https://github.com/PlayersCommittee/gemp-swccg-public/pull/1086 , which already Closes #176 with the same converted-location restore. Closing this PR so there is one PR per issue. Sorry for the extra review load.

PR #1150, opened 2026-09-17
Friendly Fire does not target the count-set to be lost

### LINKED ISSUE #447: Oh, Switch Off - Protected 6 Droids During Battle
Opponent battled with 7 characters, one Nute Gunray and the rest droids. I played Alter/Friendly Fire. He used Oh, Switch Off. All droids were protected from all weapons for the duration of the battle. If this is not a bug, I think an errata is in order.

https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=Jmwerner$rlyz3ngtf351h1ch

billbisco:
**Why**

#447: Alter / Friendly Fire into Nute + droids, then Oh, Switch Off, protected every droid for the rest of the battle. Engine marked the whole character set `TO_BE_LOST` during Friendly Fire initiation, so Oh, Switch Off canceled the interrupt and applied rest-of-turn protection to every droid in that set.

**What it does**

Friendly Fire and Alter & Friendly Fire (V) no longer target the counted characters to be lost at initiation. After destiny succeeds, opponent chooses one to be lost; that nested targeting is `TO_BE_LOST`. `ChooseCardsToLoseFromTableEffect` keeps the card owner's performing player, so opponent-choice is still an attempt by the player of Friendly Fire. Oh, Switch Off cancels that lose and protects only the chosen droid.

**What it fixes**

Oh, Switch Off is not legal when Friendly Fire is first played. It is legal when a droid is chosen to be lost. Only that droid gets rest-of-turn stolen / hit / lost protection.

**Issues:** Closes #447


BOTVHD:
If all of your characters are protected from being lost (due to previous plays of Oh, Switch Off, for example), this would still allow Friendly Fire to be played and target your group of characters (that can't be targeted to be lost)?

Beyond that, it looks like this would still force you to lose one of them (even though they should all be protected...)?

billbisco:
Friendly Fire can still be played. Its play condition is two of opponent's characters and a weapon at the site, not a character that can currently be lost.

It does not target those characters to be lost when played, so Oh, Switch Off does not answer the play.

After destiny succeeds, opponent chooses one character to be lost. That choice is targeting to be lost. A droid already protected by Oh, Switch Off cannot be chosen. If every character there is protected, none are lost.

BOTVHD:
> After destiny succeeds, opponent chooses one character to be lost. That choice is targeting to be lost. A droid already protected by Oh, Switch Off cannot be chosen. If every character there is protected, none are lost.

Hmm, where does that 'none are lost' get enforced?  With the removal of the TargetingReason.TO_BE_LOST checking in the front end of Friendly Fire, I don't see where anything prevents the player from being forced to choose one of the characters (even though they are all protected).  What restricts the character selection (for character to choose to be lost) to only ones that can be targeted to be lost?

billbisco:
Thank you, VHD.

The lose choice is ChooseCardToLoseFromTableEffect, which always adds TargetingReason.TO_BE_LOST. TargetCardsOnTableEffect then ANDs Filters.canBeTargetedBy(..., TO_BE_LOST), so a character already protected by Oh, Switch Off is not a legal choice.

If every character there is protected, that targeting finds fewer than the minimum of 1 and returns without choosing. None are lost. The count-set at initiation is still just opponent's characters at the site (play condition), not a TO_BE_LOST target set.

PR #1152, opened 2026-09-17
Enclosed vehicle immunity protects characters aboard (Closes #439)

### LINKED ISSUE #439: Rogue 1 with Commander Luke did not have immunity, even with Echo Base Garrison
In two separate battles, Commander Luke piloting Rogue 1 did not have immunity to attrition. These were on turns 5 and 6. Additionally, Master Kenobi did not let me search for a related battleground on turn 5. 

The replay can be found here: https://gemp.starwarsccg.org/gemp-swccg/game.html?replayId=Wend$fvbw197r84z02s1d

Thanks for all you do!

billbisco:
Why: leftover #439. Printed enclosed-vehicle rule: characters aboard are protected by that vehicle's immunity to attrition. Commander Luke piloting enclosed Rogue 1 with Echo Base Garrison was forced to forfeit against 1 attrition.

What it Does: Characters aboard an enclosed vehicle share that vehicle's immunity to attrition. They remain optional attrition forfeits when the vehicle is immune; they are not removed from the forfeit list. Adds Card_3_066_Tests and Card_3_155_Tests: Blizzard 2 passenger shares immunity < 4; with attrition and no battle damage that passenger and Blizzard 2 stay optional choices and Dark can decline. Punishing One with Dengar piloting stays immune < 3; a Stormtrooper passenger can still be forfeited.

What it Fixes: Enclosed-vehicle passengers share the vehicle's immunity query. They are not required to pay attrition the vehicle is immune to, and they are still allowed as optional forfeits.

Issues: Closes #439


billbisco:
Thank you, VHD. Writing the play-the-game test so it fails on unfixed code first was the right call.

The first change only made a Blizzard 2 passenger report immunity < 4. They were still a required forfeit against attrition 1. A test written only after that patch never reached Dark's forfeit prompt, so it stayed green either way.

The test now fails before the fix (Stormtrooper passenger is offered to pay attrition 1, battle damage 0) and passes after the forfeit list skips enclosed-vehicle characters who share covering immunity. That second change is what a player actually sees.

Bug tests will be written this way going forward: fail on current code first, then the same test must pass after the fix.

billbisco:
Thank you, VHD. You are right. Passengers stay optional forfeits. The forfeit-list skip is reverted. The Blizzard 2 test now asserts optional choice and decline.

billbisco:
Closing this PR. Remaining attrition is ignored using cards present at the location. Characters aboard an enclosed vehicle are not present there, so they do not need the vehicle's immunity number. The original Rogue 1 force-forfeit is still #439; this share does not address it.

